PR #6, opened 2019-05-09
BEP-6: Add delist proposal

HaoyangLiu:

PR #5, opened 2019-05-03
BEP-5: Token Resources [IMPROVEMENT]

jpthor:
This BEP describes an improvement to the BEP-2 Token standard to add resources for token management on the Binance Chain.

Token Resources are managed by a URI that is specified as part of an issued token on Binance Chain.

PR #4, opened 2019-05-01
BEP4: 256-bit Tokens

wjmelements:
#### Changes
This adds a proposal to increase the number of bits for tokens from 63 to 256.
Reviewers @notatestuser @huangsuyu

PR #3, opened 2019-04-21
BEP3: Add HTLC

caffeinum:
# BEP-3: Add HTLC to Binance Chain

- [BEP-3: Add HTLC to Binance Chain](#bep-3-htlc-to-binance-chain)
  * [1.  Summary](#1--summary)
  * [2.  Abstract](#2--abstract)
  * [3.  Status](#3--status)
  * [4.  Motivation](#4--motivation)
  * [5.  Specification](#5--specification)
    + [5.1 HTLC contract](#53-HTLC-contract)
  * [6. License](#6-license)

## 1.  Summary

This BEP describes a proposal for add HTLC to the Binance Chain.

## 2.  Abstract

BEP-2 Proposal describes a common set of rules for correct token/cryptocurrency swap from another blockchain to Binance Chain

## 3.  Status

WIP 

## 4.  Motivation

New assets can be issued on Binance Chain or come from another blockchains. During a swap, users are forced to use additional human-controlled centralized systems (issuer) for get his assets on binance chain instead old. It threats to reverse the important gains of decentralization.

### 5.1 HTLC Contract

https://github.com/bitcoin/bips/blob/master/bip-0065.mediawiki#trustless-payments-for-publishing-data

## 6. License

The content is licensed under [CC0](https://creativecommons.org/publicdomain/zero/1.0/).

PR #7, opened 2019-05-23
BEP 7: Bring Non-fungible Tokens(NFTs) to Binance Chain and expand the use cases.

amazingandyyy:
bep-7

With Non-fungible Tokens, we can expand the token use cases on Binance Chain to be:

- Event Ticket
- VIP Privilege Status
- Collectibles Assets
- more ...

Every token is transferrable, tradable, mintable, burnable and most important, non-fungible.


amazingandyyy:
# BEP-7: Non-fungible Tokens on Binance Chain

- [BEP-7: Non-fungible Tokens on Binance Chain](#bep-7-non-fungible-tokens-on-binance-chain)
  - [1.  Summary](#1--summary)
  - [2.  Abstract](#2--abstract)
  - [3.  Status](#3--status)
  - [4.  Motivation](#4--motivation)
  - [5.  Specification](#5--specification)
    - [5.1 Token Properties](#51-token-properties)
      - [5.1.1 Metadata Example](#511-metadata-example)
    - [5.2 Token Management Operation](#52-token-management-operation)
      - [5.2.1 Issue token](#521-issue-token)
      - [5.2.2 Transfer Tokens](#522-transfer-tokens)
      - [5.2.3  Freeze Tokens](#523--freeze-tokens)
      - [5.2.4  Unfreeze Tokens](#524--unfreeze-tokens)
      - [5.2.5 Mint Tokens](#525-mint-tokens)
      - [5.2.6 Burn Tokens](#526-burn-tokens)
  - [6. Token Use Case Examples](#6-token-use-case-examples)
  - [7. License](#6-license)

## 1.  Summary

This BEP describes a proposal for non-fungible tokens(NFTs) management on the Binance Chain.

## 2.  Abstract

BEP-7 Proposal describes a common set of rules for non-fungible token management within the Binance Chain ecosystem. It introduces the following details of a non-fungible token on Binance Chain:

- What information makes a non-fungible token on Binance Chain
- What actions can be performed on a non-fungible token on Binance Chain

## 3.  Status

This BEP is under drafting.

## 4.  Motivation

Design and issue non-fungible asset on the Binance Chain, as the basic economic foundations of the blockchain.

## 5.  Specification

### 5.1 Token Properties

- Source Address: Source Address is the owner of the issued token.

- Token Name: Token Name represents the long name of the token - e.g. "MyToken".

- Symbol: Symbol is the identifier of the newly issued token.

- Total Supply: Total supply will be the total number of issued tokens.

- Mintable: Mintable means whether this token can be minted in the future, which would increase the total supply of the token

- Metadata(OPTIONAL): Contains more details about the assets which this NFTs represent

### 5.1.1 Metadata Example

This is the example of what the "Metadata" can contain

```json
{
  "description": {
    "type": "string",
    "description": "Describes the asset to which this NFT represents"
  },
  "image": {
    "type": "string",
    "description": "A URI pointing to a resource with mime type image/* representing the asset to which this NFT represents. Consider making any images at a width between 320 and 1080 pixels and aspect ratio between 1.91:1 and 4:5 inclusive."
  }
}
```

### 5.2 Token Management Operation

#### 5.2.1 Issue token

Issuing token is to create a new non-fungible token on Binance Chain. The new non-fungible token represents ownership of something new, and can also peg to existing tokens from any other blockchains.

**Data Structure for Issue Operation**: A data structure is needed to represent the new token:

| **Field**    | **Type** | **Description**                                              |
| :------------ | :-------- | :------------------------------------------------------------ |
| Name         | string   | Name of the newly issued asset, limited to 32 unicode characters,  e.g. "ABCcoin" |
| Symbol       | string   | The length of the string for representing this asset is between 8 and 16 alphanumeric characters and is case insensitive. ".B" suffixed symbol is also allowed for pegging to those tokens already exist on other chains. The symbol is suffixed with the first 3 bytes of the issue transaction hash to remove a constraint of requiring unique token names. The native token, BNB, does not require this suffix. |
| Total Supply | int64    | The total supply for this token can have a maximum of 8 digits of decimal and is boosted by 1e8 in order to store as int32. The amount before boosting should not exceed 90 billion. |
| Owner        | Address  | The initial issuer of this token, the BNB balance of issuer should be more than the fee for issuing tokens |
| Mintable     | Boolean  | Whether this token could be minted(increased) after the initial issuing |
| Metadata     | object   | OPTIONAL. Contains more details about the assets which this NFTs represent |

The data in all the above fields are not changeable after the Issue Transaction, except “Total Supply” can be changed via “Mint” or “Burn” operations.

**Symbol Convention:**

[Symbol][.B]-[Suffix]

Explanations: Suffix is the first 3 bytes of the issue transaction’s hash. It helps to remove the constraint of requiring unique token names. If this token pegs to an existing blockchain, there should be an additional suffix of “.B”.

**Issue Process:**

- Issuer signed an issue transaction and make it broadcasted to one of Binance Chain nodes
- This Binance Chain node will check this transaction. If there is no error, then this transaction will be broadcasted to other Binance Chain nodes
- Issue transaction is committed on the blockchain by block proposer
- Validators will verify the constraints on total supply and symbol and deduct the fee from issuer’s account
- New token’s symbol is generated based on the transaction hash. It is added to the issuer’s address and token info is saved on the Binance Chain

#### 5.2.2 Transfer Tokens

Transfer transaction is to send tokens from input addresses to output addresses.

**Message Structure for Transfer Operation**: A data structure is needed to represent the transfer operation between addresses.

| **Field** | **Type** | **Description**              |
| :--------- | :-------- | :---------------------------- |
| Input     | []Input  | A set of transaction inputs  |
| Output    | []Output | A set of transaction outputs |

**Input Data Structure:**

| **Field** | **Type** | **Description**                                              |
| :--------- | :-------- | :------------------------------------------------------------ |
| Address   | Address  | Address for token holders                                    |
| Coins     | []Coin   | A set of sorted coins, one per currency. The symbols of coins are in descending order. |

**Output Data Structure:**

| **Field** | **Type** | **Description**                                              |
| :--------- | :-------- | :------------------------------------------------------------ |
| Address   | Address  | Address for token holders                                    |
| Coins     | []Coin   | A set of sorted coins, one per currency. The denominations of coins are in descending order. |

**Coin Structure:**

| **Field** | **Type** | **Description**                                              |
| :--------- | :-------- | :------------------------------------------------------------ |
| Denom     | string   | The symbol of a token                                        |
| Amount    | int64    | The amount is positive(unsigned) and can have a maximum of 0 digits(must be an integer) of decimal and is boosted by 1e8 in order to store as int32. |

**Transfer Process:**

- Transferer initiators sign a transfer transaction and make it broadcasted to one of Binance Chain nodes
- The Binance Chain node will check this transaction. If there is no error, then this transaction will be broadcasted to other Binance Chain nodes
- Transfer transaction is committed on the blockchain by block proposer
- Validators will verify the constraints on balance. The transfer tokens and fee will be deducted from the address of the transaction initiators.
- Add the tokens to the destination addresses

#### 5.2.3  Freeze Tokens

A Binance Chain user could freeze some amount of tokens in his own address. The freeze transaction will lock his fund, thus this portion of tokens could not be used for the transactions, such as: creating orders, transferring to another account, paying fees and etc.  

**Data Structure** **for Freeze Operation**: A data structure is needed to represent the freeze operation

| **Field** | **Type** | **Description**                                              |
| :--------- | :-------- | :------------------------------------------------------------ |
| Symbol    | string   | The symbol should belong to an existing token,e.g. NNB-B90   |
| Amount    | int64    | Frozen amount for this token can have a maximum of 0 digits of decimal(must be an integer), and the value is boosted by 1e8 to store in an int32. This amount should be less than its balance and be positive|

**Freeze Process:**

- Address-holder signed a freeze transaction and make it broadcasted to one of Binance Chain nodes
- The Binance Chain node will check this transaction. If there is no error, then this transaction will be broadcasted to other Binance Chain nodes
- Freeze transaction is committed on the blockchain by block proposer
- Validators will verify the transaction initiator’s balance is no less than the frozen amount. The fee will be deducted from the transaction initiator’s address. 
- This amount of tokens in the address of the transaction initiator will be moved from balance to frozen.

#### 5.2.4  Unfreeze Tokens

Unfreezing is to unlock some of the frozen tokens in the user's account and make them liquid again.

**Data Structure** **for Unfreeze Operation**: A data structure is needed to represent the freeze/unfreeze operation

| **Field** | **Type** | **Description**                                              |
| :--------- | :-------- | :------------------------------------------------------------ |
| Symbol    | string   | The symbol should belong to an existing token,e.g. NNB-B90   |
| Amount    | int64    | The unfreeze amount can have a maximum of 0 digits of decimal(must be an integer), and the value is boosted by 1e8 to store in an int32. This amount should be no less than the frozen amount and be positve|

**Unfreeze Process:**

- Address-holder signed an unfreeze transaction and make it broadcasted to one of Binance Chain nodes
- The Binance Chain node will check this transaction. If there is no error, then this transaction will be broadcasted to other Binance Chain nodes
- Unfreeze transaction is committed on the blockchain by block proposer
- Validators will verify the transaction initiator’s frozen balance is no less than the required amount. The fee will be deducted from the address of the transaction source.
- This amount of token will be moved from frozen to balance in the transaction initiator’s address.

#### 5.2.5 Mint Tokens

Mint transaction is to increase the total supply of a mintable token. The transaction initiator must be the token owner.

**Data Structure** **for Mint Operation**: A data structure is needed to represent the mint operation

| **Field** | **Type** | **Description**                                              |
| :--------- | :-------- | :------------------------------------------------------------ |
| Symbol    | string   | The symbol should belong to an existing mintable token, e.g. NNB-B90 |
| Amount    | int64    | Added supply for this token can have a maximum of 0 digits of decimal(must be an integer), and the value is boosted by 1e8 to store in an int32. The amount before boosting operation should be less than 90 billion after mint. |

**Mint Process:**

- Token owner signs a mint transaction and makes it broadcasted to one of Binance Chain nodes
- The Binance Chain node will check this transaction. If there is no error, then this transaction will be broadcasted to other Binance Chain nodes
- Mint transaction is committed on the blockchain by block proposer
- Validators will verify the constraints on whether the token is mintable and whether the bumped total supply will pass the limit. Then increase its total supply and deduct the fee from the address of the token owner
- Newly minted tokens are added to the address of the token owner and token info is updated on the Binance Chain

#### 5.2.6 Burn Tokens

Burn transaction is to reduce the total supply of a token. The transaction initiator must be the token owner. 

**Data Structure** **for Burn Operation**: A data structure is needed to represent the burn operation

| **Field** | **Type** | **Description**                                              |
| :--------- | :-------- | :------------------------------------------------------------ |
| Symbol    | string   | The symbol should belong to an existing token,e.g. NNB-B90   |
| Amount    | int64    | Burnt supply for this token can have a maximum of 0 digits of decimal(must be an integer), and the value is boosted by 1e8 to store in an int32. The amount should be no less than its total supply |

**Burn Process:**

- Token owner signs a burn transaction and makes it broadcasted to one of Binance Chain nodes
- The Binance Chain node will check this transaction. If there is no error, then this transaction will be broadcasted to other Binance Chain nodes
- Burn transaction is committed on the blockchain by block proposer
- Validators will verify the constraints if the token’s supply is no less than the required amount, and then it decreases the total supply and deducts the fee from the address of the token owner
- Burned tokens are deducted from the address of the token owner and token info is updated on the Binance Chain

## 6. Token Use Case Examples

With Non-fungible Tokens, we can expand the token use cases on Binance Chain to be:

- Event Ticket
- VIP Privilege Status
- Collectibles Assets
- more ...

Every token is transferrable, tradable, mintable, burnable and most important, non-fungible.

## 7. License

darren-liu:
It reads to me majority of the BEP is the same as the BEP2. It will be more clear and usable to highlight the commons and differences from BEP2, which is more friendly to readers and tells details for implementation and usage.

BEP7.md
```diff
@@ -0,0 +1,268 @@
...
+- This Binance Chain node will check this transaction. If there is no error, then this transaction will be broadcasted to other Binance Chain nodes
+- Issue transaction is committed on the blockchain by block proposer
+- Validators will verify the constraints on total supply and symbol and deduct the fee from issuer’s account
+- New token’s symbol is generated based on the transaction hash. It is added to the issuer’s address and token info is saved on the Binance Chain
+
+## 6. Token Specification
+
+### 6.1 Token Properties
+
+> x/nft/types/nft.go brings metadata on-chain
```
amazingandyyy: as you can see in this [implmentation](https://github.com/cosmos/cosmos-sdk/pull/4209/files#diff-52868bd850fc32806c263c5f9de640d6), it brings metadata to cosmos on-chain.

do we want to bring it on-chain?

pros is data is on the blockchain
cons is metadata is not flexible

BEP7.md
```diff
@@ -0,0 +1,268 @@
...
+| Amount    | int64    | The unfreeze amount can have a maximum of 8 digits of decimal, and the value is boosted by 1e8 to store in an int64. This amount should be no less than the frozen amount |
+
+**Unfreeze Process:**
+
+- Address-holder signed an unfreeze transaction and make it broadcasted to one of Binance Chain nodes
+- The Binance Chain node will check this transaction. If there is no error, then this transaction will be broadcasted to other Binance Chain nodes
+- Unfreeze transaction is committed on the blockchain by block proposer
+- Validators will verify the transaction initiator’s frozen balance is no less than the required amount. The fee will be deducted from the address of the transaction source.
+- This amount of token will be moved from frozen to balance in the transaction initiator’s address.
+
```
amazingandyyy: should we add **burn token** to BEP7-NFT token?

amazingandyyy:
@darren-liu I change some contents in the past week, the biggest difference here

1. introduce the idea of `collection` which contains symbol, e.g. NNB-B90 and a []NFT nft token array)

2. every NFT has it's own incremental id and metadata field which make this NFT highly compatible to other existing NFT tokens, e.g. ERC721, to make it easy to integrate with third-party services as @opensea and then smooth out integration on @TrustWalletApp per @vikmeup's suggestion.

BEP7.md
casmirconsensys: We should include the Burn Token so as to implement across chain communication compatibility for these use cases here are creative examples of needing to invoke a Burn Function:
 
"User's could reverse the process by sending their ERC-721 Virtual Cards to the Wallet address which would then **burn** the tokens and issue an equivalent number of in game items to the user in the game" -Virtual In-game ATM... Tokeninze Second Life
https://blog.0xproject.com/the-inaugural-0x-team-hackathon-8359b68b98e7

Decentralized asset mint (color) and **Burn** (uncolor) functions asset exchange BCH using OP_RETURN -Open Assets SPV Wallets use EPOBC e.exchange color kernels (ColorCoin) BitCoin Cash.
https://www.yours.org/content/colored-coins-in-bitcoin-cash-b26804e05964/
https://news.bitcoin.com/developers-invoke-the-idea-of-bitcoin-cash-based-colored-coins/

BEP7.md
```diff
@@ -0,0 +1,320 @@
...
+## 7. Token Use Case Examples
+
+With Non-fungible Tokens, we can expand the token use cases on Binance Chain to be:
+
+- Event Ticket
+- VIP Privilege Status
+- Collectibles Assets
+- more ...
```
casmirconsensys: "Virtual ATM In-Game Video Game Dex/Asset exchange"

BEP7.md
```diff
@@ -0,0 +1,320 @@
...
+
+## 3. Status
+
+This BEP is under drafting.
+
+## 4. Motivation
+
+Design and issue non-fungible asset on the Binance Chain, as one of the basic economic foundations of the blockchain.
+
+## 5. Collection Specification
```
casmirconsensys: Chain ID List mutate state NFT HUB enforcement that enables Mutation -ZManian

BEP7.md
```diff
@@ -0,0 +1,320 @@
...
+- Validators will verify the transaction initiator’s frozen balance is no less than the required amount. The fee will be deducted from the address of the transaction source.
+- This amount of token will be moved from frozen to balance in the transaction initiator’s address.
+
+## 7. Token Use Case Examples
+
+With Non-fungible Tokens, we can expand the token use cases on Binance Chain to be:
+
+- Event Ticket
+- VIP Privilege Status
+- Collectibles Assets
```
casmirconsensys: Digital Assets Art, Game Items, rare "Jordans", token reputation ownership -vshush

BEP7.md
casmirconsensys: @oopcode suggests an infoURL in the tokenURI() metadata JSON schema encode metadata, and @okwme references cosmosSDK issues #4046 draft implementation to make it compatible with opensea's API use from coinbase NFT struct TokenID is used as a key when storing the NFT so is not contained in the NFT struct itself.

casmirconsensys: @okwnee @oopcode Billy suggests a compromise betweeen onchain and offchain data v2 separate metadata into an interface with nested features move metadata into its own module that can be used to describe any asset on a SDK chain including (fungible tokens) but specifically references NFT's Metadata/TokenURI/URL's

BEP7.md
```diff
@@ -0,0 +1,320 @@
...
+
+## 1. Summary
+
+This BEP describes a proposal for non-fungible tokens(NFTs) management on the Binance Chain.
+
+## 2. Abstract
+
+BEP-7 Proposal describes a common set of rules for non-fungible token within the Binance Chain ecosystem. It introduces the following details of a non-fungible token on Binance Chain:
+
+- What information makes a non-fungible token on Binance Chain
```
casmirconsensys: type NFT struct {
          owner            sdk.ACC.Address
`json: "owner"`
         Name             String
`json: "name"`
         Description    String
`json: "image"`
         TokenURI .     String
`json: "token_URI"`
}
@vshvsh packet goes origin->Hub->destination making a footprint in three chains move a description to the external metadata

BEP7.md
casmirconsensys: }

    struct SneakrStruct{
    bytes32 playerName;
    uint playerBadge;
    uint shoeBoost;
    uint index;
    }

BEP7.md
```diff
@@ -0,0 +1,320 @@
...
+- Token Name: Token Name represents the long name of the token - e.g. "BinanceAnnualEvent2019"
+
+- Denom: Denom is the identifier(symbol) of the newly issued collection.
+
+- Total Supply: Total supply is the total number of issued NFT tokens belong to this collection.
+
+- Mintable: A boolean value to indicate whether the owner can increase the totalSupply.
+
+- NFTs: An array contains items in NFT interface(described below).
+
```
casmirconsensys: Should we include Token Index for an array of in game token assets for a multi-verse game like ready player one?
It's included in the ERC-1155 Spec

BEP7.md
```diff
@@ -0,0 +1,320 @@
+# BEP-7: Non-fungible Tokens on Binance Chain
+
+- [BEP-7: Non-fungible Tokens on Binance Chain](#bep-7-non-fungible-tokens-on-binance-chain)
+  - [1. Summary](#1-summary)
+  - [2. Abstract](#2-abstract)
+  - [3. Status](#3-status)
+    - [Couple things to keep in mind](couple-things-to-keep-in-mind)
+  - [4. Motivation](#4-motivation)
+  - [5. Collection Specification](#5-collection-specification)
```
casmirconsensys: []NFT nft token array) is this similar to the index array from ERC-1155 and is collection the semantical version of this concept? 
Non-fungible tokens can be interacted with using an index based accessor into the contract/token data set. Therefore to access a particular token set within a mixed data contract and a particular non-fungible within that set, _id could be passed as <uint128: base token id><uint128: index of non-fungible>.

uint256 baseTokenNFT = 12345 << 128;
uint128 indexNFT = 50;

uint256 baseTokenFT = 54321 << 128;

balanceOf(baseTokenNFT, msg.sender); // Get balance of the base token for non-fungible set 12345 (this MAY be used to get balance of the user for all of this token set if the implementation wishes as a convenience).
balanceOf(baseTokenNFT + indexNFT, msg.sender); // Get balance of the token at index 50 for non-fungible set 12345 (should be 1 if user owns the individual non-fungible token or 0 if they do not).
balanceOf(baseTokenFT, msg.sender); // Get balance of the fungible base token 54321.
https://github.com/ethereum/eips/issues/1155

BEP7.md
zmanian: So there are two good ways of doing mutable NFAs in an IBC system.

The first approach requires all mutations to send IBC messages back to the issuing chain and it returns the result of the mutation.

The other solution is to have the issuer provide a set of chains whose mutations it will treat as cannonical and thus given a hub that enforces this white list would allow everyone else to treat those mutations as cannonical as well without the round trips to the issuer zone

okwme: Are the mutations we're referring to with regard to metadata or to ownership?

If wrt metadata then the whitelist scenario might result in multiple sources of truth for that metadata--informationn stored on the origin chain and information stored on the secondary chain. This would mean the origin chain would need to monitor changes to metadata instead of the secondary chain.

Since monitoring changes to metadata must take place regardless I'd suggest that it is the secondary chain who must monitor changes on the origin chain. A whitelist could be managed to determine if the secondary chain has permissions to modify aspects of the metadata on the origin chain, but for simplicity and sanity it seems that the origin chain should be the source of truth of the metadata. The secondary chain should be able to implement their own system for tracking that metadata, either keeping a local version in sync or querying the origin chain when necessary.

okwme:
> @darren-liu I change some contents in the past week, the biggest difference here
> 
> 1. introduce the idea of `collection` which contains a symbol, e.g. NNB-B90, totalSupply, and a `[]NFT` nft token array)
> 2. every NFT has it's own incremental id and metadata field which make this NFT highly compatible to other existing NFT tokens, e.g. ERC721, to make it easy to integrate with third-party services as @ProjectOpenSea and then smooth out integration on @TrustWallet per @vikmeup's suggestion.

i'm worried about the requirement of incremental id. In the spec you list ID as a string but incremental IDs make me think you imply to change to int. There are many projects which will need more fine grained control over their ID system. ERC-721 handles this well by allowing tokens to be queried by index while not requiring the token IDs to be incremental. 

For example I have a game where I encode some item information into the ID itself, this is a convenient way for me to store information about the NFT without redundant storage costs. 

Or another example might be NFTs that represent in-stock items that already have their own alphanumeric id system. It would be redundant to have multiple IDs per item and the app designer would prefer to use the same ID on both systems.

For this reason I'd suggest allowing IDs to be string but create an additional qeuerier that  requests information about denomination and the index of the NFT. Furthermore this could be extended into a full pagination with an index and a limit to access multiple NFTs incrementally while still preserving choice of ID.

BEP7.md
casmirconsensys: NFTs
IBC format for NFTs. Billy working on NFT-IBC spec. NFT will follow after the token transfer. 
Current design: NFT will not carry metadata with them. Only have identity and chain ID.
Right now the way it works in Ethereum, there is one field that is very short that can be resolved to get metadata. Billy is wondering whether to continue with that approach, or have the metadata on chain. The ability to update the metadata should live with the zone, rather than transfer over IBC. When you are transferring NFTs over IBC, you would only transfer the denomination - you know where the origin chain is, and you can go a query that chain for the metadata.
Zaki: Being able to have a white list of multiple zones that are allowed to mutate the metadata is going to be valuable. Instead of saying, just transfer the ID, you can transfer the metadata along with a white list of zones that have the ability to change the metadata. Hub can enforce that only metadata updates from white listed chains are persistent.
A use case for the above: Breeding  crypto kitties example. If the metadata can only be changed by issuing chain, then need to return the kitty every time to make a change to it. A white list would allow other zones to also breed crypto kitties.
NFT spaces being driven by gaming. Financial derivative NFTs should have a different spec. The same spec may not suit gaming and financial NFTs.
Question 1 from Meher: How is the whitelist made In the case of NFT staking derivatives where the issuer is the protocol?
Question 2 from Meher: If multiple genes can mutate the metadata of an NFT, and many of these mutations happen for the same NFT on different chains, won’t the trustworthiness of the final state of the NFT be dependent on these security of all of these chains? and this dependence would imply that it security is very low?
Zaki: Small number of zones that can make changes to the NFT metadata.  this is like a contract between Chains,  where the issuing chain issues the right to mutate metadata to a few number of chains that are near it in the IBC graph. Issuing zone can effectively debase a NFT tokens, thus always keeping the power to recover from bad mutations.
The issuing chain is pretty special and lots of things have to go back to that. 
There are also around ideas such that: Original issuance of NFTs can be someone other zone, but the hub ensures some kind of security properties of the NFTs.
These designs leading to a world where there is an authoritative source, and other somewhat less authoritative sources. As a user you can figure out the relationship, and refer back to the authoritative Zone when needed. There may be some advantages of working with the less authoritative version, as it might have better transaction properties or better connectivity.
Decision: Sum all this up on the NFT ICS issue.

okwme: Yes I agree that there should be the option for external chains to update the metadata on the origin chain. A white list would be a convenient method but I think there may be more complicated systems needed by some chains and it should be up to the origin chain to implement as simple or as complicated of a system as they need. A white list would be a great solution for simple cases and might be enough for Binance Chain.

okwme:
For reference here is a summary spec of the Cosmos-NFT Module:
https://hackmd.io/@okwme/cosmos-nft

BEP7.md
```diff
@@ -0,0 +1,320 @@
...
+
+- NFTs: An array contains items in NFT interface(described below).
+
+```go
+// Collection of non fungible tokens
+type Collection struct {
+  Owner         string  `json:"owner"`                    // owner of the collection;
+  Name          string  `json:"name"`                     // name of the collection;
+  Denom         string  `json:"denom,string,omitempty"`   // the identifier(symbol) of the newly issued collection;
+  TotalSupply   int64   `json:"totalSupply"`              // total supply of the NFT in this collection;
```
okwme: Is `TotalSupply` ever going to be different from the length of the array of `NFTs`?

BEP7.md
```diff
@@ -0,0 +1,320 @@
...
+  Mintable      bool    `json:"mintable"`                 // can increase total supply or not;
```
okwme: Is this adequate? Doesn't it also need information about who can mint etc? Is it relevant only for BEP native NFTs?

fedekunze:
> For reference here is a summary spec of the Cosmos-NFT Module:
https://hackmd.io/@okwme/cosmos-nft

Moved to https://github.com/cosmos/cosmos-sdk/pull/4766

BEP7.md
casmirconsensys: @amazingandyyy is there a particular rationale for TotalSupply in64? I know that ERC 1155 gives you the option of minting 1 as a unique NFT or an array for limited "collections". I think that over-specification, in the spec could be confusing as the ability to mint on chain would be meant to allow custom specific use cases whereby an NFT would not nessarily need to declare a total supply unless it was required by an open sea type wallet. Now we could include the specs to give that capability as standard specific to BEP7 that would allow for others to return to our spec as the most cross chain combatible as Loom Network and Wink share cross chain capabilities with Smart Contract side chains that are Ethereum based, or EOS, or TRON.

BEP7.md
casmirconsensys: Yes, this is only relevant to BEP native NFT's @amazingandyyy I think he specified this within the BEP2 spec for native BNB tokens as they were ERC-20 but were made to be specific to the Binance Chain. 
I think that the approach used in Smart Contract ERC-721 and ERC-1155 is to use inheritance to delegate control of the Ownership Contract, so maybe perhaps the structure of the Ownership Contract would be in the Mint Type function and thus making Ownership a parameter that is always set to Owned by origin chain and as ownership is transferred between chain A and chain B then the host(origin) chain can delegate that Ownership responsibility to the receiving chain to then Transact or Trade on its behalf and have all the functionality of the Ownership chain except for the TotalSupply. This could be part of the on-chain metadata packet that gets sent along with Ownership permissions. In a BEP native chain, this would allow ultimate cross-chain reliability of that metadata to be read by any wallet that can read permissions. They could go freely between Smart Contract based wallets as they would have met the criteria to be traded for ERC-1155 or ERC-721 and in some cases where need be they could present themselves as BEP-2 ERC-20 tokens for efficiency.

BEP7.md
casmirconsensys: NFT Module intergrated with the Cosmos SDK v.0 v.1 
-Metadata Interface 
-Metadata Struct includes the interface for offchain key-value pairs
-Introspection -description JSON that confirms the Attribute(s)

IBC Standard
-ID includes tokenID
-Denom includes demonination

IBC NFT Transfer x origin chain
-ID tokenID
-originChain B.originChain 
-Metadata on wheter or not it sold on Binance Dex
-approval vs transfer in regards to cross-chain transfer of ownership delegation of utility in regards to its readability by wallets regarding its fungibility status during that time period whether it is meant to be a BEP2/ERC-20 token or semi-fungible ERC-1155 in such case that it is a collection of ERC-721 unique one of a kind NFT or an (index)NFTarray[]



ERC-1155 interoperability with Enjin Coin Unity SDK

dGoods Mythical Goods EOS 
-Blockchain Game developers 
-Schema.org Web3 W3c off-chain consortium (metadata)
-github

BEP7.md
amazingandyyy: @casmirconsensys I am not sure what do you mean **over-specification** , `int64` simply means it can be any number.

amazingandyyy: @okwme  yes, for example, totalySupply is 1M, and the collection starts from zero token in it, then owner can mint new NFT into the []NFTs away until it hits 1M cap.

BEP7.md
amazingandyyy: @okwme that is about implmentation, as it says in the mint token [operation](https://github.com/binance-chain/BEPs/pull/7/files/bafb62d95b4d455d6bc41f755316524b5b91dc7a#diff-03fa5f1a99f35738a833285aee92fa25R189): 

```
Issue token

 This can be performed only when the following conditions are satisfied:
 1. Executor is the collection owner.
 2. The amount of the minted tokens on the collections has not exceeded the total supply of the collection.
```

BEP7.md
amazingandyyy: `[]NFT` is golang's implementation, what you are referring to is ethereum's implementation, it is very much different since it's in solidity. The spec here is discussing tech spec, scope only to  
1. what fields
2. what actions can be performed

BEP7.md
amazingandyyy: yea, but here is talking about `token` use case, not really about what applications, for example, 
Virtual ATM, In-Game Video Game or Dex/Asset exchange all counted under collectible assets

BEP7.md
amazingandyyy: We need to discuss what the Metadata should be, right now it's a URI referring to an external resource.

BEP7.md
amazingandyyy: You mean chainId? what is the Token index is referring to here?

BEP7.md
casmirconsensys: Thanks @amazingandyyy for clarification on this golang vs sol tech spec I wanted to frame a reference point since I am coming from an Etherem Solidity background.

BEP7.md
casmirconsensys: Yeah, the call tonight is talking with @okwme about Application Specifics in regards to the BEP7

BEP7.md
casmirconsensys: Yeah, @amazingandyyy chainID, this was from the last call last week @okwme mentioned the ERC-1155 Spec in the call.

BEP7.md
casmirconsensys: > @casmirconsensys I am not sure what do you mean **over-specification** , `int64` simply means it can be any number.

This was a question @okwme raised during the call from last week so we wanted to get clarification on what that meant thanks for clearing that up! @amazingandyyy

amazingandyyy:
@casmirconsensys I can see one major thing is unclear is about the chianId, may need to have eng from binance come in and give their idea of cross-chain/IBC/cosmos...etc technical details on this.

BEP7.md
amazingandyyy: yes, I myself was also confused by solc's implementation before.

BEP7.md
amazingandyyy: cool, yea not sure how binance plan to achieve cross-chain in the IBC/cosmos ecosystem in the future

BEP7.md
amazingandyyy: I see, yea, I hope my explanation clarify it.

okwme:
download

amazingandyyy:
Summary from zoom call with cosmos & binance team on 7/31 12pm: 
1. cosmos will release BasicNFT keeper(including collection/nft) in a seperated modules repo, and hope binance can build another keeper layer(binance-specific keeper) on top of it
2. for metadata concern, `tokenURI` is only properties in the BasicNFT keeper
3. The scope of this PR(BEP7) is to propose of 
   a. what binance-specific keeper(more properties) potentially can do
   b. binance-specific handler(more management) potentially can do

BEP7.md
```diff
@@ -0,0 +1,320 @@
...
+#### 6.2.3 Freeze Tokens
+
+A Binance Chain user could freeze some amount of tokens in his own address. The freeze transaction will lock his fund, thus this portion of tokens could not be used for the transactions, such as: creating orders, transferring to another account, paying fees and etc.  
+
+**Data Structure** **for Freeze Operation**: A data structure is needed to represent the freeze operation
+
+| **Field** | **Type** | **Description**                                                 |
+| :--------- | :-------- | :------------------------------------------------------------ |
+| Denom     | string   | The symbol of a collection - e.g. NNB-B90                       |
+| ID        | string   | The id of a token in the collection                             |
```
HaoyangLiu: Do you think the freeze period is necessary? If there is no freeze period, the owners can freeze and unfreeze their assets at anytime, thus the freeze operation will have no limitation for owners.

BEP7.md
```diff
@@ -0,0 +1,320 @@
...
+This can be performed only when the following conditions are satisfied:
+
+1. Executor is the collection owner.
+
+2. Mintable value of the collection is true
+
+| **Field**  | **Type**  | **Description**              |
+| :--------- | :-------- | :---------------------------- |
+| Denom      | string    | The symbol of the collection |
+| Amount     | int64     | The amount is positive and can have a maximum of 8 digits of decimal and is boosted by 1e8 in order to store as int64. |
```
HaoyangLiu: Is it necessary to have decimals for non-fungible tokens?

amazingandyyy:
yes, @casmirconsensys it will be great if this can get merged and the community can move forward to help

casmirconsensys:
@amazingandyyy Let's Do This!! Who else is on board to approve this thing? @okwme @zmanian

casmirconsensys:
> @casmirconsensys I can see one major thing is unclear is about the chianId, may need to have eng from binance come in and give their idea of cross-chain/IBC/cosmos...etc technical details on this.

What have we heard from Binance Eng about this so far? @amazingandyyy

okwme:
confused why my comments are marked as "pending" still...
Screen Shot 2020-05-14 at 11 18 16
Screen Shot 2020-05-14 at 11 18 23

BEP7.md
```diff
@@ -0,0 +1,320 @@
+# BEP-7: Non-fungible Tokens on Binance Chain
+
+- [BEP-7: Non-fungible Tokens on Binance Chain](#bep-7-non-fungible-tokens-on-binance-chain)
```
amazingandyyy, line 3: testing

casmirconsensys (approved):
Would be nice to have some XML-JSX-TSX Metadata Schemas added.

brilliant-lx:
outdate, archive it.

PR #11, opened 2019-05-30
BEP-11: IEOs on Binance Chain

hiddentao:
# BEP-11: IEOs on Binance Chain

## 1.  Summary

This BEP describes a proposal for conducting Initial Exchange Offerings (IEOs) on the Binance Chain.

## 2.  Abstract

A specification for running IEOs directly on Binance Chain. This specification attempts to provide details on how the currently active IEO process (as of May 29, 2019) can be transplanted to run on Binance Chain, including any necessary modifications to the chain protocol itself.

## 3.  Status

This BEP is in Draft status.

## 4.  Motivation

Several IEOs have been successfully conducted on Binance.com. However, they lack the benefits of transparency and decentralization which Binance Chain provide. Furthermore, it seems that most (if not all) IEOs (Harmony.One) result in contributors being awarded BEP-2 tokens that are native to Binance Chain. Thus it makes sense to take things one step forward and conduct the IEO itself on Binance Chain, thus providing far greater transparency and auditability to the process.

In summation, here are some of the benefits:

* Greater transparency and auditability, thereby increasing trust in the IEO process, and thus its popularity.
* No need to bridge from Binance.com to Binance Chain (i.e. BNB no longer needs to be sent to Binance.com in order to participate, thus reducing participant risk).
* Faster turnaround time from IEO to tradeability of crowdfunded token, since the offered token is already present on Binance Chain.

Read more -> https://github.com/hiddentao/BEPs/blob/7de7fabb9a33f1f413bd3c1a3dae4f5fa8807ef1/BEP11.md

### DIFF
--- a/BEP11.md
+++ b/BEP11.md
@@ -7,6 +7,15 @@
   * [3.  Status](#3--status)
   * [4.  Motivation](#4--motivation)
   * [5.  Specification](#5--specification)
+    + [5.1 Creating the Offering](#51-creating-the-offering)
+    + [5.2 Editing the Offering](#52-editing-the-offering)
+    + [5.3 Claim period - claiming tokens](#53-claim-period-claiming-tokens)  
+    + [5.4 Lottery allocation](#54-lottery-allocation)  
+    + [5.5 Withdrawing tokens](#55-withdrawing-tokens)  
+    + [5.6 Ending an Offering](#56-ending-an-offering)  
+    + [5.7 Cancelling an Offering](#57-cancelling-an-offering)  
+    + [5.8 Random value generator](#58-random-value-generator)
+    + [5.9 KYC & Sybil resistance](#59-kyc-and-sybil-resistance)  
   * [6. License](#6-license)
 
 ## 1.  Summary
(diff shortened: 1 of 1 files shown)


darren-liu:
Personally I think the on-chain random number generation and also the KYC require more details to work on.

hiddentao:
@darren-liu Yep, I'm still giving these some thought.

* For the random number generator, one method would be to implement the well-known [RANDAO Commit-Reveal algorithm](https://github.com/randao/randao) with the [no-duplicates fix](https://revelry.co/critical-randao-vulnerability/). Alternatively, a more attractive option would be to use BLS signatures to generate the random number. I know the Cosmos team are already looking into potentially integrating BLS signatures for validation itself, perhaps this can be re-used for an RNG as well.

* Regarding KYC, the main requirement is a concept of "global storage" in Binance Chain, which can be written to by certain addresses, but read by anyone. I need to look into Tendermine more to determine how this can be made feasible.

touhonoob:
> Regarding KYC, the main requirement is a concept of "global storage" in Binance Chain, which can be written to by certain addresses, but read by anyone. I need to look into Tendermine more to determine how this can be made feasible.

Wouldn't that cause privacy issues like leaking passport photos?

hiddentao:
@touhonoob If you read my original draft, I propose storing a "token" of sorts against a given Binance Chain address to represent KYC approval. The token would consist of:

```
kyc_approved: <Boolean> 
country: <ISO 2-letter code>
```

Thus, for a given item in that list, you would only know which country the address is registered in, and whether it is KYC approved. The actual documents would reside with Binance.com, same as they are currently are.

hiddentao:
I've posed a question to the Cosmos team regarding RNGs -> https://forum.cosmos.network/t/generating-cryprographically-secure-random-numbers/2270

hiddentao:
The DGaming team have implemented BLS signature-based random beacon for Cosmos here -> https://github.com/dgamingfoundation/tendermint/blob/dcr-random/docs/arcade/arcade.md

brilliant-lx:
outdated, archived it.

PR #17, opened 2019-06-14
[R4R]BEP12: Introduce customized scripts and transfer memo validation

HaoyangLiu:
## Background 
For tokens on binance chain, recently exchanges tender to use a single address for client deposit and require clients to input memo to identify client account information. 

This causes a lot of problems for both exchange customer support and clients. Clients may be panic when their tokens are missing and exchange supports have to verify the client deposit transactions manually. This is caused by that many users don’t realize that memo is required or they just forget to do that. What’s worse, sometimes their wallets don’t support transaction memo at all.

So we plan to implement a memo check mechanism on binance chain to filter out all transactions which are intend to deposit tokens to exchanges and have no valid memo.

## Motivation
This memo validation can be used for any membership based deposit/charge business model.

This BEP also proposes an important infrastructure for customized scripts. In the future, more amazing features will be built on it.

### DIFF
--- a/BEP12.md
+++ b/BEP12.md
@@ -1,0 +1,118 @@
+# BEP12: Implement Memo Validation
+## Summary
+This BEP describes a new feature that enables users to customize additional memo validation.
+## Abstract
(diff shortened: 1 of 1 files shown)


jonesk7734 (approved):
😀

PR #18, opened 2019-06-14
[R4R] BEP-18: for state sync enhancement

ackratos:
### Background

[State sync](https://docs.binance.org/fullnode.html#state-sync) is a way to help newly-joined user sync latest status of binance chain. It sync latest syncable peer's status so that fullnode user (who want catch up with chain as soon as possible) doesn't need sync from block height 0 (with cost that discard all historical blocks locally).

 BEP-18 Proposal describes an enhancement of existing state sync implementation to improve user experience. It introduces the following details:

 - What's the procedure to take a snapshot
- What's the procedure to sync snapshot from other peers
- Snapshot (manifest, snapshot chunks) format

### Motivation

We propose this BEP to enhance full node user experience (and ease their paining) on using state sync because of following implementation limitation.

 1. Users complain most about state syncing testnet is very slow and usually stucking on some request. <br> <br> In this enhancement, we want data responsed more evenly across peers so that syncing can continously making progress and the overall syncing time can reduce from 30 - 45 min to around 5 min.

 2. Interruption during state sync (node process get killed because of reboot computer or user inpatient) would make already synced state in vain (because current full node doesn't persist synced part on dist) more worse it mistakenly write a lockfile prevent user state sync again. <br> <br> In this enhancement, we want support break-resume downloading and keep consistent status for arbitrarily restart.

 3. User cannot control which height they can sync from. Current implementation only allows user sync latest syncable height. <br> <br> In this enhancement, we want make syncing from historical syncable height possible. User either specify a height to sync in config file or follow latest syncable height.


BEP18.md
```diff
@@ -0,0 +1,123 @@
...
+### 5.1 Take snapshot
+
+There are two ways to take snapshots from a fullnode: automatically or manually. Snapshots will be put under `$HOME/data/snapshot/<height>`. All types involved in the snapshot are encoded by go-amino and compressed by snappy. More details will be explained later.
+
+1. To make fullnode automatically take snapshots, just make sure `state_sync_reactor` in `$HOME/config.toml` is set to true.
+2. To manually take snapshots, stop the node if it is running, then run `./bnbchaind snapshot --home <home> --height <height>`.
+
+If the snapshot taking procedure is interrupted, the node will be still in good status, but it cannot provide the interrupted height for other peers.
+
+Note: automatically take snapshot would keep occupying disk space. Fullnode would not delete them automatically, so the user should periodically delete unneeded snapshots manually if they want to save disk space.
```
melekes: What prevents you from deleting them automatically?

BEP18.md
```diff
@@ -0,0 +1,123 @@
...
+
+### 5.2 Sync snapshot
+
+Syncing snapshot is designed to be only run once during full node first start. To enable state sync from others, `state_sync_reactor` should be true and `state_sync_height` should be set to non-negative (default `-1` means disable syncing from others).
+
+If a user wants to sync from (majority) peers' latest sync-able height, they should set `state_sync_height` to 0.
+
+Stop and restart fullnode during state sync is allowed. The next time full node is started, it will resume by loading Manifest and downloaded chunks then download missing chunks.
+
+Once state sync is successful, a `STATESYNC.LOCK` file will be created under `$HOME/data` to prevent state sync next time.
```
melekes: Does this mean node can only sync once?

BEP18.md
```diff
@@ -0,0 +1,123 @@
...
+
+### 5.3 Manifest format
+
+Manifest serves as a summary of snapshot chunks to be synced. It also maintains the order and types of snapshot chunks. Fullnode firstly asks peer's for the manifest file at the beginning of state sync and will trust majority peers with the same manifest.
+
+SHA256 hash sum of each chunk synced will be checked against the hash declared within the manifest file.
```
melekes: Does it make sense to hash the manifest file itself and give the user ability to provide that hash when it wants to sync to a specific height?

BEP18.md
```diff
@@ -0,0 +1,123 @@
...
+| AppStateHashes | []SHA256Sum | hashes of app state chunks                                                                                                                                                                                                                                                                                                                                       |
+| BlockHashes    | []SHA256Sum | hashes of the block in this snapshot, currently only one block is synced                                                                                                                                                                                                                                                                                             |
+| NumKeys        | []int64     | number of keys for each sub-store, this sacrifices clear boundary between cosmos and tendermint, saying tendermint knows application DB might be organized by sub-stores.<br><br> |
```
melekes: What if the order or number of sub-stores changes? Or this is controlled by Version (snapshot version)?

BEP18.md
```diff
@@ -0,0 +1,123 @@
...
+#### 5.4.1 App state chunk
+
+ App state chunk includes iavl tree nodes. Usually, each app state chunk takes up to 4MB serialized iavl tree nodes (before snappy compression). 
+ 
+ Iavl tree node bigger than 4MB is split into different incomplete chunks, that's where `Completeness` field effect.
+
+| Field        | Type     | Description                                                                                                                                                                                  |
+|--------------|----------|----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| StartIdx     | int64    | compare (startIdx and number of complete nodes) against (Manifest.NumKeys) we can know each node's sub-store                                                                                      |
+| Completeness | uint8    | flag of completeness of this chunk, not enum because of go-amino doesn't support enum encoding                                                                                               |
```
melekes: can't this be a `bool`?

BEP18.md
ackratos: > What prevents you from deleting them automatically?

The frequency of snapshot taken is once per day for binance chain. And in our testnet, it takes 300M per snapshot and should be around just 30M at production considered the account numbers is ten times less than testnet there. Whole year less than 15GB storage cost is not a concern for now.

BEP18.md
ackratos: > Does this mean node can only sync once?

nodes are supposed to only sync once. But as I wrote in this BEP, they can delete this file manually and trigger state sync again if they are too many days lagged. (but their `blockchain.db` is not continuous in this case)

BEP18.md
ackratos: > Does it make sense to hash the manifest file itself and give the user the ability to provide that hash when it wants to sync to a specific height?

actually, the user can set `state_sync_height` in `config.toml`to the snapshot height other peers have. But the height number is not fixed for every day, we are considering how to let node knowns what snapshot heights I can provide.

The height is much simpler than the hash of manifest right?

BEP18.md
ackratos: > What if the order or number of sub-stores changes? Or this is controlled by Version (snapshot version)?

State sync will accept majority peer's manifest. Saying if the order or length of `NumKeys` of manifest in some nodes is inconsistent with the majority node (maybe they forget to upgrade their binary whereas others changed the order or num of sub stores), the new joined node won't sync from it (accept his manifest file).

BEP18.md
ackratos: > can't this be a `bool`?

It has 4 kind of value: 0 complete, 1 - in complete start, 2 - in complete middle, 3 - in complete end.
I have left more detailed explaination on `StartIdx` field, you can read that:)

darren-liu: please add these values into the explanation.

BEP18.md
melekes: Yeah, but height + hash might give the user additional insurance (like when the nodes the user trust publish height + height of the snapshot and the user uses them to sync the state)

ankitamadan:
I am able to sync from any historical syncable height on or after 10th July 2019 but I am not able to sync from snapshot height before 10th July 2019
E.g
I can sync from any height on or after this block height -> `19060575 (10th July 2019)`
I cannot sync from height on or before this block height -> `18829401 (9th July 2019)`

PR #20, opened 2019-06-21
BEP-20: Issuer Controlled Token Transfer Whitelist

aquabu:
This BEP describes a proposal for adding and removing addresses from a token issuer controlled whitelist to restrict who can receive tokens.

Whitelists controlled by security token owners can enable regulatory compliance and significantly improve liquidity for private markets.

To enable this Binance Chain would need to implement a whitelist of addresses maintained by the owner of the contract - the issuer. Only investors who have passed the issuers AML/KYC and accredited investor checks would be added to the whitelist for unrestricted trading. Additional checks would continue to be performed by individual exchanges who are accountable for their own vetting. More complex transfer restrictions such as maximum investment amounts for everyday investors are out of scope for this proposal.

### DIFF
--- a/BEP-20.md
+++ b/BEP-20.md
@@ -1,0 +1,84 @@
+# BEP-20: Issuer Controlled Token Transfer Whitelist
+
+  - [1. Summary](#1-summary)
+  - [2. Abstract](#2-abstract)
+  - [3. Status](#3-status)
+  - [4. Motivation](#4-motivation)
+  - [5. Specification](#5-specification)
+  - [6. References](#6-references)
+  - [7. License](#7-license)
+
+## 1.  Summary 
+
+This BEP describes a proposal for adding and removing addresses from a token issuer controlled whitelist to restrict who can receive tokens. 
+
+_This is not investment or legal advice._
+
+## 2.  Abstract
+
+Whitelists controlled by security token owners can enable regulatory compliance and significantly improve liquidity for private markets. 
+
+To enable this Binance Chain would need to implement a whitelist of addresses maintained by the owner of the contract - the issuer. Only investors who have passed the issuers AML/KYC and accredited investor checks would be added to the whitelist for unrestricted trading. Additional checks would continue to be performed by individual exchanges who are accountable for their own vetting. More complex transfer restrictions such as maximum investment amounts for everyday investors are out of scope for this proposal.
+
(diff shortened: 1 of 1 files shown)


darren-liu:
Is this `restricted token` supposed to be listed and traded within the limited address scope?

aquabu:
@darren-liu yes. The restricted token is to be traded only between addresses in the limited address scope - referred to as the "whitelist".

hiddentao:
Where would the whitelist data be stored?

HaoyangLiu:
On Binance Dex, if the controlled tokens are used to pay order fee, then the controlled tokens will be paid to the validators. We have to take this case into consideration.

BEP-20.md
```diff
@@ -0,0 +1,84 @@
...
+
+## 5.  Specification
+
+### 5.1 Token Attributes For Whitelist
+
+* Tokens have a `whitelist_required_for_transfer` attribute of type `boolean`
+* Tokens have a `whitelist` of type `[address]`
+
+### 5.2 Token Whitelist Attribute Access Restrictions
+* Only the token owner can set the `whitelist_required_for_transfer` boolean value
```
HaoyangLiu: A token owner might lose his private key. If this happens, the whitelist can be changed any more.  Do you think it is necessary to enable a token owner to specify more manage addresses which can change the whitelist too?

aquabu: Yes, great point. This would be a serious problem. I think the solution for this is native mult-signature support for binance chain private keys. Maybe this goes in a new BEP? 

In the mean time token owners would need to find a native threshold signature or key sharding solution for secp256k1 keys. If there is a multi-sig wallet solution like this available for Binance Chain then that may be the initial recommended best practice.

HaoyangLiu:
I'm sorry for that we plan to close this PR and you can create another PR for your proposal. As you know, [EIP20](https://eips.ethereum.org/EIPS/eip-20) proposed a token stardand ERC20 on Ethereum. We also want to use 20 to represent our token stardand. So the BEP20 is reserved for the token stardand on Binance Smart Chain: https://github.com/binance-chain/BEPs/pull/75.

aquabu:
@HaoyangLiu No problem. This is from a while ago and I know there are alternate solutions now. Thanks!

PR #23, opened 2019-07-01
Propose BEP23 Continuous WalletConnect

darren-liu:
Wallet Connect is a great protocol to ease the use of DApp and make Web Wallet much safer. However, it is not designed for trading and brings a few inconveniences for traders:
- The connection is a short term of the transient. Once the mobile phone app switches to the background, e.g. disturbed by a phone or screen timeout, the connection will be cut off. The users “unlocked” wallet would be locked again and cannot perform any actions.
- Even the connection is there, every order or cancel will require traders to unlock the mobile phone (after inputting password or touch/face ID), and click the button, which is quite time-consuming in a fast ticking market.

Here a longer, continuous-time period is required as a “secure slot” for traders to send, and cancel orders freely and safely.

This proposal is to ask for enhancement on both DApp and mobile wallet side to get a better user experience meanwhile less security sacrifice.

### DIFF
--- a/BEP23.md
+++ b/BEP23.md
@@ -69,7 +69,7 @@ In order to not to blindly and silently sign the transactions, users can enable 
 
 ### 5.3 Web Wallet Continuous Connection
 Even the mobile wallets can keep the connection with Bridge Server, it also requires the DApp side (Web Wallet) can manage to connect with Bridge Server on the same topic. As users may often refresh the pages to resolve trading API issues or trading page instability. After each time the page is refreshed, the DApp should:
-look up the topic in the localsession storage, and connect to the Bridge Server with the same topic
+look up the topic in the localStorage, and connect to the Bridge Server with the same topic
 after the re-connection is established, confirm with Bridge Server whether mobile wallet is still connected. If the mobile wallet is not connected any more (through several times of check, to ensure the disconnection is there), DApp may stop connecting and ask for user to re-establish the whole process. This is to prevent mobile wallet connection is lost, and this is a new part of communication not included in the official Wallet Connect protocol. 
 
 ## 6. License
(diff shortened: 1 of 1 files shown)


BEP23.md
```diff
@@ -0,0 +1,78 @@
...
+```
+1. Is it still within any “Timeout Window”? If yes, go to step 2; otherwise prompt for user consent;
+2. Is it a NewOrder or Cancel transaction, if yes, go to step 3 or 4; otherwise prompt for user consent;
+3. If it is NewOrder, if the base asset of the trading pair is one of “Allowed Symbols” and order quantity is less than or equal to “Allowed Order Size” and total order quantity is smaller than “Allowed Total Quantity” and the number of orders should be smaller than “Allowed Number of Orders”, sign the transaction and refresh the time window if “Refreshable” is true; otherwise, prompt for user consent;
+4. If it is NewOrder, if the base asset of the trading pair is one of “Allowed Symbols” and order quantity is less than or equal to “Allowed Order Size” and total order quantity is smaller than “Allowed Total Quantity” and the number of orders should be smaller than “Allowed Number of Orders”, or “CancelIsRelaxed” is true, sign the transaction and refresh the time window if “Refreshable” is true; otherwise, prompt for user consent;
```
huangsuyu: If it is Cancel order?

notatestuser:
I would suggest adding a ping message, sent from wallet to client side, to check that the app is still open.  We should disconnect (show RED connection status) if the client does not respond to this ping request after a certain amount of time.

This would prevent an issue where the app is closed/disconnected but the user still sees a connected status symbol and is still able to queue transactions.

Thanks to @huangsuyu's QA efforts :)

brilliant-lx:
outdated, archive it.

blaaauw:
Hi everyone.


Please I use samsung Ao3 stop all transactions or withdrawal.


Thanks
***@***.***
0740380373

PR #38, opened 2019-07-22
BEP38: Non-interactive Multi-Signature Wallet.

canyateam:
# Non-interactive Multi-Signature Wallet. 

This BEP describes a new feature that builds on the [multi-signature module](https://hub.cosmos.network/docs/gaiacli.html#multisig-transactions) that is part of the Cosmos Gaia mainnet. 
The essence of this feature is that there is no interactivity required in the use of the multi-signature, and is closer in function to the [Etheruem Gnosis Multi-signature](https://github.com/gnosis/MultiSigWallet). 

## Overview
The non-interactive version of this described in this BEP instead uses Blockchain key-value store to store a basic amount of state that outlines the multi-signature account. 
The result of this is that parties do not need to cooperate or interact and instead can sign the wallet independently and at different times. 
Additionally state-less wallet interfaces can be built that connect to Binance Chain with nothing but a connected wallet. 

This is a substantially better user experience, as shown in this demo video:

[DEMO VIDEO](https://www.dropbox.com/s/s0c6mcor8l5c0ym/Multisig.m4v?dl=0)

[SOURCE CODE](https://github.com/cbarraford/cosmos-multisig)

## Motivation

While offering a far better user experience, this BEP is important because it lays the groundwork for the following modules that are under design by the CanYa RnD team:

- **Escrow Module**: Escrow funds between parties
- **Hedged Escrow Module** Escrow and hedge funds between parties. 

These modules are important for increased utility and useability of Binance Chain.

### DIFF
--- a/BEP38.md
+++ b/BEP38.md
@@ -1,0 +1,337 @@
+# BEP-38: Non-interactive Multi-Signature Wallet
+
+- BEP-38: Non-interactive Multi-Signature Wallet
+  - [1. Summary](#1-summary)
+  - [2. Abstract](#2-abstract)
+  - [3. Status](#3-status)
+  - [4. Motivation](#4-motivation)
+  - [5. Specification](#5-specification)
+    - [5.1 CLI](#51-cli)
+    - [5.2 API](#52-api)
+  - [6. Developer Notes](#6-developer-notes)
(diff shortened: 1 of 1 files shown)


brilliant-lx:
outdated, archive it.

PR #39, opened 2019-07-23
BEP-39: Add MEMO to Transfer WebSocket.

canyateam:
# BEP-39: Add MEMO to Transfer WebSocket.

BEP-39 requests that `MEMO` data field be added to the `/ws/userAddress` [websocket](https://docs.binance.org/api-reference/dex-api/ws-streams.html#3-transfer).

Currently the `MEMO` field is not being returned on the websocket, which means that services that rely on MEMO to set transaction specifications must then retrieve it from the [Transaction API endpoint](https://docs.binance.org/api-reference/dex-api/paths.html#transaction).

This creates unneccessary burden on the API and slows down the transaction processing.

The solution is to add it to the Transfer Websocket as a data field to stream.

### DIFF
--- a/BEP39.md
+++ b/BEP39.md
@@ -1,0 +1,65 @@
+# BEP-39: Add MEMO to Transfer WebSocket.
+
+- BEP-39: Add MEMO to Transfer WebSocket
+  - [1. Summary](#1-summary)
+  - [2. Abstract](#2-abstract)
+  - [3. Status](#3-status)
+  - [4. Motivation](#4-motivation)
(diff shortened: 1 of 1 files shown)


ackratos:
@canyateam this has been deployed.

Thana11:
1409900311986

Thana11:
https://www.paypal.me/Kuu4411
4150288429  krungthai​ bank​  (THB)​

jonesk7734 (approved):
😀

PR #41, opened 2019-07-28
Whitelist Script for BEP12

wjmelements:
#### Changes
Transactions can send money to the wrong address, causing immediate loss of value.
A whitelist could be useful for a wallet or service to protect against mistake transfers, as an additional safeguard.
This also proposes transactions for adding to the whitelist or removing from it.
Reviewers @huangsuyu

### DIFF
--- a/BEP12.md
+++ b/BEP12.md
@@ -44,7 +44,7 @@ By default, all accounts’ flags are zero which means no script is specified. S
 
 Besides, the account flags changes will take effect since the next transaction.
 
-### Memo Check Script for Transfer
+### 0x1: Memo Check Script for Transfer
 This script is aimed to ensure the transfer transactions have valid memo if the receivers require this.
 
 Firstly, this script will check the following conditions:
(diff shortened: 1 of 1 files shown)


wjmelements:
I pushed a commit to make the parameters plural, which I see as an improvement, though it may complicate the fee structure. `git revert be37795` if unwanted.

#### Proposed fee structure
To account for the increase in state size, adding whitelisted addresses should cost more per-account. However, because removing whitelisted accounts trims the state size, removing blacklisted accounts should have a smaller, flat fee.

BEP12.md
```diff
@@ -72,6 +72,59 @@ return nil
...
+    }
+    return nil
+}
+```
+
+
+#### New Transaction: WhitelistRecipients
+A node should reject the transaction if any of the accounts are already whitelisted by this sender.
+
+Parameters for adding whitelisted recipients:
```
huangsuyu: We need to allocate extra space for whitelist storage. It will be a scability issue.

wjmelements:
Overhead would depend on the node implementation. If the account is
amino-encoded, for example, there is no extra space for null fields.

HaoyangLiu:
> Overhead would depend on the node implementation. If the account is amino-encoded, for example, there is no extra space for null fields.

@huangsuyu means that this feature requires not only a validation script but also extra storage for saving whitelist.

wjmelements:
> > Overhead would depend on the node implementation. If the account is amino-encoded, for example, there is no extra space for null fields.
> 
> @huangsuyu means that this feature requires not only a validation script but also extra storage for saving whitelist.

added

darren-liu:
why cannot this be implemented on client side, i.e. wallets or signing service? They are more flexible to add more features, e.g. different group of white list can accept different size of fund, or an address book that can associate address with monikers to select.

Storing on chain cost token and actually exposes more privacy information of the addresses relationship. I don't think many people are interested.

brilliant-lx:
outdate, archive it.

PR #64, opened 2019-12-13
BEP-64: BNB Payment URI Scheme

BNB48ClubSirIan:
Aim to build the standard of URI for asset transferring on binance-chain.
Once adopted by binance chain clients, it could make transactions on binance-chain more smooth.

### DIFF
--- a/BEP_draft_URI.mediawiki
+++ b/BEP_draft_URI.mediawiki
@@ -38,18 +38,18 @@ Elements of the query component may contain characters outside the valid range. 
 
 (See also [[#Simpler syntax|a simpler representation of syntax]])
 
- binanceurn     = "binance:" binanceaddress [ "?" binanceparams ]
+ binanceurn     = "bnb:" binanceaddress [ "?" binanceparams ]
  binanceaddress = *base58
  binanceparams  = binanceparam [ "&" binanceparams ]
- binanceparam   = [ amountparam / memoparam / assetparam]
+ binanceparam   = [ amountparam / memoparam / assetparam / metadata]
 
  amountparam    = "amount=" *digit [ "." *digit ]
  memoparam     = "memo=" *qchar
  assetparam   = "asset=" *qchar
+ metadata = <optional, base64 encoded 1024 maximum raw bytes>
+Here, "qchar" corresponds to valid characters of an RFC 3986 URI query component, excluding the "=" and "&" characters, which this BEP takes as separators. And optional metadata field is for extensible purpose, customizable content for different app parsing.
 
-Here, "qchar" corresponds to valid characters of an RFC 3986 URI query component, excluding the "=" and "&" characters, which this BEP takes as separators.
-
-The scheme component ("binance:") is case-insensitive, and implementations must accept any combination of uppercase and lowercase letters. The rest of the URI is case-sensitive, including the query parameter keys.
+The scheme component ("bnb:") is case-insensitive, and implementations must accept any combination of uppercase and lowercase letters. The rest of the URI is case-sensitive, including the query parameter keys.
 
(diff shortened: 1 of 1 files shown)


xiaoguang-binance:
Payment is great use case for Binance Chain.  Some feedbacks: 
1. URI Scheme:
In bitcoin: It already includes URI extensions to cover merchant validation, https://github.com/bitcoin/bips/blob/master/bip-0072.mediawiki. One example: bitcoin:mq7se9wy2egettFxPbmn99cK8v5AFq55Lx?amount=0.11&r=https://merchant.com/pay.php?h%3D2a8628fc2fbe
In ethereum, it introduces prefix for different use cases (e.g., pay-) . https://github.com/ethereum/EIPs/blob/master/EIPS/eip-831.md
Maybe, we can include these two extensions in this proposal. 

2. Payment Protocol: payment protocol is critical to guide the wallet implementation. Maybe, we can create a separate BEP as https://github.com/bitcoin/bips/blob/master/bip-0070.mediawiki.

BEP_draft_URI.mediawiki
```diff
@@ -0,0 +1,128 @@
...
+
+(See also [[#Simpler syntax|a simpler representation of syntax]])
+
+ binanceurn     = "binance:" binanceaddress [ "?" binanceparams ]
+ binanceaddress = *base58
+ binanceparams  = binanceparam [ "&" binanceparams ]
+ binanceparam   = [ amountparam / memoparam / assetparam]
+
+ amountparam    = "amount=" *digit [ "." *digit ]
+ memoparam     = "memo=" *qchar
```
ackratos, line 47: IMHO the memo is not enough for the merchant. An extendible metadata field (base64 encoded 1024 maximum raw bytes for different app parsing) might be helpful. Typical usage:
1. A wallet app want to protect the privacy of recipients, so they generated a fake address in URL but generated an encrypted address that is included in the metadata field. Only wallet app can decrypt this address and make the transfer
2. Merchant want add description of product i.e. "Coca-Cola 250 ml"

BEP_draft_URI.mediawiki
```diff
@@ -0,0 +1,128 @@
...
+
+Binance URIs follow the general format for URIs as set forth in RFC 3986. The path component consists of a binance address, and the query component provides additional payment options.
+
+Elements of the query component may contain characters outside the valid range. These must first be encoded according to UTF-8, and then each octet of the corresponding UTF-8 sequence must be percent-encoded as described in RFC 3986.
+
+=== ABNF grammar ===
+
+(See also [[#Simpler syntax|a simpler representation of syntax]])
+
+ binanceurn     = "binance:" binanceaddress [ "?" binanceparams ]
```
ackratos, line 41: I suggest use `bnb:` or `bncpayment`

BNB48ClubSirIan: Update to `bnb:` in latest commit.

BEP_draft_URI.mediawiki
BNB48ClubSirIan: Can't agree with this more.

huangsuyu:
Will this support `multisend` tx?

ackratos:
> Will this support `multisend` tx?

discussed with DJ&xg&@sophie offline, we might have the requirement for social payment, but currently, we can handle it by metadata field rather than expand the `address` field to an array

huangsuyu:
> > Will this support `multisend` tx?
> 
> discussed with DJ&xg&@sophie offline, we might have the requirement for social payment, but currently, we can handle it by metadata field rather than expand the `address` field to an array

we can support multi-send with multiple inputs, in case your balance is insufficient under only one account

PR #67, opened 2020-03-13
BEP-67: Price-based Order Expiration

erhenglu:
This BEP describes an enhancement of the Order Expiration.  
Currently orders on Binance Chain will be expired after 3 days. It is better if we could keep some open orders longer so that traders do not need re-create the expired orders every 3 days.  

The solution is to keep orders in the best 500 price level for 30 days rather than 3 days.

### DIFF
--- a/BEP67.md
+++ b/BEP67.md
@@ -1,0 +1,47 @@
+# BEP-67: Price-based Order Expiration
+
+- [BEP-67: Price-based Order Expiration](#bep-67-Price-based-order-expiration)
+  - [1. Summary](#1-summary)
+  - [2. Abstract](#2-abstract)
+  - [3. Status](#3-status)
+  - [4. Motivation](#4-motivation)
+  - [5. Specification](#5-specification)
+    - [5.1 Order Expiration](#51-order-expiration)
+    - [5.2 Change Impact](#52-change-impact)
+  - [6. License](#6-license)
+
+## 1. Summary 
+
+This BEP describes an enhancement of the [Order Expiration](https://docs.binance.org/faq.html#what-is-order-expire).  
+
+## 2. Abstract
+
+Currently orders on Binance Chain will be expired after 3 days. Order cannot live long on the market even their price stays competitive, which is not convenient and incurs cost to traders. The solution is to keep orders in the best 500 price levels for 30 days rather than 3 days.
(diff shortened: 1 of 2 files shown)


BEP67.md
```diff
@@ -0,0 +1,48 @@
...
+    - [5.2 Change Impact](#52-change-impact)
+  - [6. License](#6-license)
+
+## 1. Summary 
+
+This BEP describes an enhancement of the [Order Expiration](https://docs.binance.org/faq.html#what-is-order-expire).  
+
+## 2. Abstract
+
+Currently orders on Binance Chain will be expired after 72 hours. This may lead to high frequency of order placement, which is not very convenient to users.
```
darren-liu, line 19: not exactly 72 hours. "high frequency" is a word often used in "high frequency trading", which means trade multiple  time in a sub-second time period. I suggest we say:

"Order cannot live long on the market even their price stays competitive, which is not convenient and incur cost to traders."

BEP67.md
```diff
@@ -0,0 +1,48 @@
...
+
+## 4. Motivation
+
+Some users want to keep their open orders more than 3 days so that they do not need re-create their cancelled orders every three days. 
+
+
+## 5. Specification
+
+###  5.1 Order Expiration
+In the 1st block after UTC 00:00 every day, orders which have been staying in order book for longer than 72 hours will be removed from order book and marked as 'expired'.
```
darren-liu, line 34: please highlight this is the "current implementation".

BEP67.md
```diff
@@ -0,0 +1,47 @@
...
+    - [5.2 Change Impact](#52-change-impact)
+  - [6. License](#6-license)
+
+## 1. Summary 
+
+This BEP describes an enhancement of the [Order Expiration](https://docs.binance.org/faq.html#what-is-order-expire).  
+
+## 2. Abstract
+
+Currently orders on Binance Chain will be expired after 3 days. Order cannot live long on the market even their price stays competitive, which is not convenient and incur cost to traders. The solution is to keep orders in the best 500 price level for 30 days rather than 3 days.
```
darren-liu, line 19: => incurs cost 
=> 500 price levels

jonesk7734 (approved):
😀

PR #70, opened 2020-04-08
BEP70: List and Trade BUSD Pairs

forcodedancing:
BUSD is one of the most popular stable coins in the world, and people are using it to exchange fiat currencies everyday. If token owners can list BUSD trading paris directly, without dependency on BNB trading pairs, on Binance Chain, it will definitely benefit the markets, token owners and traders. 

BEP70 proposes a solution for listing and trading BUSD pairs without explicit dependency on BNB.

With BEP70, we would like to achieve: 
1) Relaxing the restriction of listing, and 
2) Easy channels to exchange BUSD stable coins


jonesk7734 (approved):
😀

PR #81, opened 2020-11-21
BEP-81: Tokenized Fee Service Layer For Binance Chain

centerprime:
# BEP-81: Tokenized Fee Service Layer For Binance Chain

## 1.  Summary

This BEP describes a proposal for executing smart contract functions by paying Tokens instead of BNB on the Binance Chain.

## 2.  Abstract

This BEP describes a proposal to improve the properties of BEP20 on Binance chain to pay tokens for transactions instead of gas (BNB) to execute the smart contract function.

## 3.  Status

This BEP is already implemented.

## 4.  Motivation

If a first-party token owner generates a transaction by improving the properties of BEP20 on Binance chain, the BEP delegates the Transaction Gas Fee (BNB) to a third party to transfer the token, paying for the gas and receiving a fee in token.
The owner who issued the token prefers to form the token economic and utility by paying transaction fees with the issued token. Current BEP20 tokens cannot pay fees with token itself, so token holders must have a separate gas (BNB). In order to improve this problem, if the token owner is referred to as a third party, the token owner pays gas cost (BNB) and the third party receives the token.
The structure above is the same as the exchange of classical finance, and the token infrastructure can be analyzed through the exchange rate.
CenterPrime Project's following proposal is to propose a BEP for “Token Infrastructure Analysis via Token Return Rate”.

### DIFF
--- a/BEP71.md
+++ b/BEP71.md
@@ -1,0 +1,120 @@
+
+# BEP-71: Tokenized Fee Service Layer For Binance Chain
+
+- [Tokenized Fee Service Layer For Binance Chain](#bep-71-smart-contract-execution-fee)
+  - [1. Summary](#1-summary)
+  - [2. Abstract](#2-abstract)
+  - [3. Status](#3-status)
+  - [4. Motivation](#4-motivation)
+  - [5. Specification](#5-specification)
+    - [5.1 Transactions](#51-transactions)
+      - [5.1.1 How it works](#511-confirmation_transaction)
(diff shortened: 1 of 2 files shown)


centerprime:
Hello Binance Chain team,
We have uploaded BEP81 source code to our Github page.
https://github.com/centerprime/SmartContract/tree/master/bep81

brilliant-lx:
outdated, archive it.
And this BEP seems a bit like part of the abstract account, which is under our schedule to consider, not the roadmap is not clear right now.

PR #101, opened 2021-04-08
BEP91: Increase Block Gas Ceiling for Binance Smart Chain

HaoyangLiu:
# BEP-91: Increase Block Gas Ceiling for Binance Smart Chain

- [BEP-91: Increase Block Gas Ceiling for Binance Smart Chain](#bep-91-increase-block-gas-ceiling-for-binance-smart-chain)
  - [1. Summary](#1-summary)
  - [2. Abstract](#2-abstract)
  - [3. Status](#3-status)
  - [6. License](#6-license)
  
## 1. Summary

This BEP describes a proposal for the increment of **gasceil** of BSC.

## 2. Abstract

BEP-91 Proposal describes a change of BSC node configuration, and suggests Validator operators should increase the value of  the block `GasCeil`from `40000000` to `60000000`.

## 3. Status

This BEP is a draft. 

## 4. Motivation

The increasing adoption of BSC leads to a more active network. Blocks start hitting the gasceil daily. The increasing prevalence of blocks near the limit may lead to increased fees and delays in the processing of transactions in the future. This will undermine the core value of BSC as a fast and cheap network. 

img

BSC network has once reached the 40M gas limit, it can handle more transactions according to the following screenshot.

img

Increased gasceil will lead to more transactions per block. It will be a direct solution to scale.  

## 5. Specification

According to the past laboratory tests, BSC can process real transactions with more than &70M Gas. Considering the complex network environment in mainnet and the different hardware of validators, 60M is a relatively reasonable value. If 60M is larger than expected, the miner module of BSC will adaptively adjust it to avoid network overloading. 

It is strongly recommended to raise the gasceil parameter as well as hardware. The validator may get less reward than expected if its hardware fails to validate enough tx.

As the transaction volume continues to increase, more changes should be made to the internal logic on transaction execution, storage and P2P communication.

## 6. License

### DIFF
--- a/BEP91.md
+++ b/BEP91.md
@@ -1,0 +1,45 @@
+# BEP-91: Increase Block Gas Ceiling for Binance Smart Chain
+
+- [BEP-91: Increase Block Gas Ceiling for Binance Smart Chain](#bep-91-increase-block-gas-ceiling-for-binance-smart-chain)
+  - [1. Summary](#1-summary)
+  - [2. Abstract](#2-abstract)
+  - [3. Status](#3-status)
+  - [4. Motivation](#4-motivation)
+  - [5. Specification](#5-specification)
+  - [6. License](#6-license)
+  
+## 1. Summary
+
+This BEP describes a proposal for the increment of **gasceil** of BSC.
+
+## 2. Abstract
+
+BEP-91 Proposal describes a change of BSC node configuration, and suggests Validator operators should increase the value of  the block `GasCeil`from `40000000` to `60000000`.
+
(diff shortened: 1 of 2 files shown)


VenTiLaT:
I'm sorry but what it means ? More fees for users of BSC ?
BNB price increase every day, I hope that Binance will not make the same mistake than Ethereum network !
I think that users wait for deflation of fees when BNB get value...
Juggernaut @Fakk2 (Twitter).

darren-liu (approved):

PR #111, opened 2021-07-27
BEP91: raise gaslimit from 70MM to 100MM

guagualvcha:
Recently, the BSC Blocks start hitting the gasceil daily:
image

The increasing prevalence of blocks near the limit may lead to increased fees and delays in the processing of transactions in the future. 

I propose to raise the gas limit from 75MM to 100MM to relieve the network congestion.


CruiseMan:
Is this still relevant ?

PR #128, opened 2022-01-07
[R4R] BEP-128: Improvement on BNB Smart Chain Staking Reward Distribution

forcodedancing:
# BEP-128: Improvement on BNB Smart Chain Staking Reward Distribution

- [BEP-128: Improvement on BNB Smart Chain Staking Reward Distribution](#bep-128-improvement-on-bnb-smart-chain-staking-reward-distribution)
  - [1. Summary](#1-summary)
  - [2. Abstract](#2-abstract)
  - [3. Status](#3-status)
    - [5.1 Overall Workflow](#51-overall-workflow)
    - [5.2 Distribution Batch Size](#52-distribution-batch-size)
    - [5.3 User Impact](#53-user-impact)
  - [6. License](#6-license)

## 1. Summary
This BEP describes a proposal to improve the BNB Smart Chain staking reward distribution mechanism on BNB Beacon Chain.

## 2. Abstract
Instead of distributing BNB Smart Chain staking rewards in a single block for each day, this proposal suggests distributing staking rewards in many consecutive blocks, to minimize the burden on the specific block.

## 3. Status
This BEP is a draft.

## 4. Motivation
Currently, the staking rewards of BNB Smart Chain are distributed in the first block of a day on BNB Beacon Chain (which is known as breath block). With the increasing number of delegators (more than 50,000 nowadays), it will lead to a heavy load to breath block, and users' transactions/requests could be affected. Meanwhile, it could also be a bottleneck for further evaluation of BNB Beacon Chain. Thus, this proposal provides a solution to fix the issue and benefit the evaluation of BNB Beacon Chain in the long term.

### 5.1 Overall Workflow
image

The overall distribution processes work like this:
 - In breath block, validators' commission fee will be distributed, all delegators' rewards will be saved to a separate store.
 - Delegators' reward will be distributed in batches in the following blocks.

### 5.2 Distribution Batch Size

Distribution batch size controls how many delegators will receive reward in a single block. It should be evaluated and adjusted to make sure:
 - The reward distribution must finish within one day, which means the batch size should not be too small.
 - It should not be too big to bring heavy load on a single block.

At the same time, this batch size can also be governed.

### 5.3 User Impact

The impact to general delegators is minimal, that their rewards will be delayed a bit after applying this proposal. However, the delay is relatively small and can be ignored compared to the whole staking process. Taking the current delegation volume as an example, about 50,000 rewards are distributed, then in the worst case the delay is 3~4 minutes (500 blocks) if the batch size is 100.

## 6. License


BEP128.md
```diff
@@ -0,0 +1,48 @@
...
+
+## 3. Status
+This BEP is a draft.
+
+## 4. Motivation
+Currently, BSC’s staking rewards are distributed in the first block of a day on Binance Chain (which is named breath block). With the increasing number of delegators (more than 50,000 nowadays), it will lead to a heavy load to breath block, and users' transactions/requests could be affected. Meanwhile, it could also be a bottleneck for further evaluation of Binance Chain. Thus, this proposal provides a solution to fix the issue and benefit the evaluation of Binance Chain in the long term.
+
+## 5. Specification
+### 5.1 Overall Workflow
+![overall workflow](https://user-images.githubusercontent.com/61674316/148506577-aac9d05a-488d-4ff0-b879-1aa1a0e5364c.png)
```
yutianwu, line 28: can we create an asset folder to place the images instead of referring the images outside

yutianwu: like this https://github.com/ethereum/EIPs/tree/master/assets

forcodedancing: sure

guagualvcha (approved):

yutianwu (approved):

Debu976116 (commented):
STA

PR #189, opened 2023-01-13
Bep lockable tokens proposal

Zionodes:
This is a BEP for the lockable tokens extension standard suitable for ERC20 and ERC777 tokens. It allows to lock the tokens used for lending instead of transfer.


Rubens1969:
***@***.***

jonesk7734 (approved):
😀

PR #301, opened 2023-10-20
BEP-301: validators can produce consecutive blocks

NathanBSC:
<pre>
  BEP: 301
  Title: Validators can produce consecutive blocks
  Status: Draft
  Type: Standards
  Created: 2023-09-27
  Discussions(optional): https://forum.bnbchain.org/t/bep-idea-validators-can-produce-consecutive-blocks/2052
</pre>

# BEP-301: Validators can produce consecutive blocks

- [BEP-301: Validators can produce consecutive blocks](#bep-301-validators-can-produce-consecutive-blocks)
  - [1. Summary](#1-summary)
  - [2. Abstract](#2-abstract)
    - [4.1 Scaling Principle](#41-scaling-principle)
    - [4.2 Implementation Specification](#42-implementation-specification)
      - [4.2.1 Priority Allocation](#421-priority-allocation)
      - [4.2.2 Validator Set Switch](#422-validator-set-switch)
      - [4.2.3 Block Avoidance](#424-block-avoidance)
      - [4.2.4 Governable Number of Consecutive Blocks](#425-governable-number-of-consecutive-blocks)
  - [5. Incentive Fairness Analysis](#5-incentive-fairness-analysis)
  - [6. Security Analysis](#6-security-analysis)
  - [7. License](#7-license)

## 1. Summary 
Each epoch in BSC consists of multiple slots, and a batch of validators take turns in a predefined order to obtain priority block-producing rights for each slot. This BEP proposes an adjustment to the allocation of priority block-producing rights: each validator receives priority block-producing rights for a predetermined number of consecutive slots per round.
## 2. Abstract 
BEP-301 describes a new allocation method for priority block-producing rights, which is logically concise and significantly enhances the system's transaction processing capacity, while remaining orthogonal to transaction processing optimization techniques within blocks.
## 3. Motivation 
The BSC ecosystem is active and continuously evolving, requiring ongoing improvements to the system's transaction processing capacity.
### 4.1 Scaling Principle 
Currently, each validator obtains priority block-producing rights for a single slot and is then rotated, with a fixed block interval of t. The transaction processing limit is t/2 for validating transactions from the previous block and t/2 for processing transactions in the new block.
<div align="center">

</div>
Assuming the number of transactions that can be processed is T, the average TPS is calculated as follows:
<div align="center">

</div>
After implementing this BEP, each validator can obtain priority block-producing rights for a continuous sequence of n slots in one round. The first block they process still allocates t/2 for validating transactions from the previous block and t/2 for processing transactions in the new block. However, subsequent blocks can skip the transaction validation process and focus on processing transactions in the new block. It is recommended that the processing time for the last block in their current round be only half of the time required to process transactions in the new block. This helps avoid the situation where the first block processed by the next validator is close to being an empty block.
<div align="center">

</div>
The TPS after continuous block production is as follows:
<div align="center">

</div>
Therefore, the TPS improvement rate is:
<div align="center">

</div>
The relationship between the number of continuous blocks produced and TPS improvement is 
<div align="center">

</div>
As shown in the graph, it is evident that continuous block production cannot double the TPS. There is no benefit when n<3, and increasing n beyond 5 does not proportionally increase the benefits. Therefore, a value within the range [3, 5] is suitable for n; when n=4, r=50%.

#### 4.2.1 Priority Allocation 
Each epoch predefines a set of validators, with a total of validatorN validators, and each validator within the set has a unique index ranging from [0, validatorNum). If the current block height is blockN, then the validators with the following indices obtain priority block-producing rights.
<div align="center">

</div>

#### 4.2.2 Validator Set Switch 
Each epoch will choose a new validator set, assuming an epoch contains epochSlots slots. The validator set switch occurs only when the block height reaches Bswitch to prevent epoch block forging. The calculation of Bswitch is as follows:
<div align="center">

</div>
In addition, the rewards for FastFinality are also distributed when there is a transition in the validator set.

#### 4.2.3 Block Avoidance 
To prevent fewer than 1/2 of the nodes from controlling the entire network, block producers are required to produce fewer than n blocks within the previous validatorN*n/2 historical blocks.

#### 4.2.4 Governable Number of Consecutive Blocks 
When n=1, it is equivalent to disabling the feature of consecutive block production, while significant optimization is observed when n belongs to the range [3,5]. Currently, the range for the value of n is set to [1,9] but except 2.

## 5. Incentive Fairness Analysis
Within a single epoch, tail validators have fewer block-producing opportunities, but the allocation of priority rights is unbiased and cannot be manipulated. Therefore, from a statistical perspective, it is fair.
## 6. Security Analysis
This BEP relies on BSC's Fast Finality feature. If Fast Finality fails, it may result in the following issues:
1. Nodes intentionally hide mined blocks, potentially leading to longer short-term reorganizations. 
2. The probability of finality for transactions increases, requiring waiting for 2/3 validatorN * n + blocks. 

In addition, as the consecutive time during which a single validator gains priority in block production increases, it may make MEV extraction easier.

## 7. License 


buddh0:
**Whether it's version v0.1 or v0.2, it cannot prevent some validators from collaborating to abandon Fast Finality votes in order to gain benefits and harm the network.**

palafoxbeta (approved):

NathanBSC:
for simplicify, we can cancel dynamical gas limit

PR #336, opened 2023-12-08
Implement EIP-4844: Shard Blob Transactions

emailtovamos:


BEPs/BEP-336.md
```diff
@@ -0,0 +1,108 @@
...
+### 6.1 Blob pool
+Once a blob transaction is sent, it ends up in the blob pool which is a separate pool only for blob transactions.
+
+### 6.2 Verification / Filtering
+The blob transaction undergoes initial verification just like in case of a normal transaction.
+There's however an extra verification step in `validateBlobSidecar()`. This should verify proofs of the blobs using `VerifyBlobProof()`. 
+Once the verification is passed the transaction is stored in the blobpool storage.
+
+### 6.3 Consensus process: Generation & Storage of Sidecars
+During consensus process, the sidecars consisting of blobs are made separately than a block. A block may have multiple sidecars. The payload of execution will include sidecars and the rest will remain the same.
```
zzzckck, line 73: The payload of execution will include sidecars and the rest will remain the same.
===
Q: what does it mean? Legacy transaction's payload will include sidecars content?

emailtovamos: I should probably change the wording. 
The point is that during consensus process, the sidecars are needed to be constructed. There are two ways of doing it. 
1 - Include the blobs etc inside the transactions themselves and fetch those during consensus directly from transactions.
2 - Transactions once added to the pool don't have blobs with them. This is what geth does to keep it lightweight. Later during consensus process, prysm separately calls to receive the blobs which are then converted to sidecars. 

In my PoC which I implemented before, I had done the first one as I mainly wanted to get it working as soon as possible. It is unclear to me if the 2nd point is necessarily the best way to go for bsc and if the performance will significantly reduce if we do the first point. 

Once I reach that stage in my current work after merging from geth, I will be in a better position.

zzzckck:
similar to: https://eips.ethereum.org/EIPS/eip-4844, we can put all 4844 stuff in one BEP, no need to create one dedicated for transaction definition.

BEPs/BEP-336.md
```diff
@@ -0,0 +1,108 @@
...
+Introduce a new transaction format for “blob-carrying transactions” which contain a large amount of data that cannot be accessed by EVM execution, but whose commitment can be accessed. The format is intended to be fully compatible with the format that will be used in full sharding.
+
+## 3. Status 
+
+Work in progress. 
+A PoC has been created in this branch: https://github.com/bnb-chain/bsc/tree/4844 
+
+## 4. Motivation 
+
+Rollups such as OpBnB offer scaling solutions of BSC. But their ability might be limited sometimes due to higher transaction fees and cost of storage in the blockchain.
```
zzzckck, line 44: OpBnB -> opBNB

BEPs/BEP-336.md
zzzckck: Blob are propagated differently from Tx:
- blob payload are broadcasted less frequently to reduce the traffic cost
- only recent blob will be kept, around 18days on Ethereum
IMHO, we need the blob pool.

BEPs/BEP-336.md
```diff
@@ -0,0 +1,501 @@
...
+Introduce a new transaction format for “blob-carrying transactions” which contain a large amount of data that cannot be accessed by EVM execution, but whose commitment can be accessed. The format is intended to be fully compatible with the format that will be used in full sharding.
+
+## 3. Status 
+
+Work in progress. 
+A PoC has been created in this branch: https://github.com/bnb-chain/bsc/tree/4844 
+
+## 4. Motivation 
+
+Rollups such as opBNB are in the short and medium term, and possibly in the long term, the only trustless scaling solution for Ethereum & BSC. Transaction fees on L1 have been very high for months and there is greater urgency in doing anything required to help facilitate an ecosystem-wide move to rollups. Rollups are significantly reducing fees for many Ethereum users: Optimism and Arbitrum frequently provide fees that are ~3-8x lower than the Ethereum base layer itself, and ZK rollups, which have better data compression and can avoid including signatures, have fees ~40-100x lower than the base layer.
```
zzzckck, line 48: `Transaction fees on L1 have been very high for months`
It is not true for BSC.

zzzckck: We may refer the EIP-4844, but need some adjustment to reflect BSC's status.

BEPs/BEP-336.md
```diff
@@ -0,0 +1,501 @@
+<pre>
+  BEP: 336
+  Title: Shard Blob Transactions
```
zzzckck, line 3: refer: https://github.com/bnb-chain/BEPs/pull/345/files
title could be: Implement EIP-4844: Shard Blob Transactions

BEPs/BEP-336.md
```diff
@@ -0,0 +1,501 @@
...
+  - [1.  Summary](#1-summary)
+  - [2.  Abstract](#2-abstract)
+  - [3.  Status](#3-status)
+  - [4.  Motivation](#4-motivation)
+  - [5.  Specification](#5-specification)
+  - [6.  Rationale](#6-rationale)
+  - [7. Backwards Compatibility](#7-backwards-compatibility)
+  - [8. Security Considerations](#8-security-considerations)
+  - [9. Life of a Blob Transaction](#9-life-of-a-blob-transaction)
+  - [9.  Life of a Blob Transaction](#9-life-of-a-blob-transaction)
```
zzzckck, line 21: duplicated.

zzzckck (approved):

mameebbu (approved):

mameebbu (approved):

PR #341, opened 2024-01-12
BEP-341: Validators can produce consecutive blocks

NathanBSC:
<pre>
  BEP: 341
  Title: Validators can produce consecutive blocks
  Status: Draft
  Type: Standards
  Created: 2023-09-27
  Discussions(optional): https://forum.bnbchain.org/t/bep-idea-validators-can-produce-consecutive-blocks/2052
</pre>

# BEP-341: Validators can produce consecutive blocks

- [BEP-341: Validators can produce consecutive blocks](#bep-341-validators-can-produce-consecutive-blocks)
  - [1. Summary](#1-summary)
  - [2. Abstract](#2-abstract)
    - [4.1 Scaling Principle](#41-scaling-principle)
    - [4.2 Implementation Specification](#42-implementation-specification)
      - [4.2.1 Priority Allocation](#421-priority-allocation)
      - [4.2.2 Validator Set Switch](#422-validator-set-switch)
      - [4.2.3 Block Avoidance](#424-block-avoidance)
      - [4.2.4 Governable Number of Consecutive Blocks](#425-governable-number-of-consecutive-blocks)
      - [4.2.5 Combatting MEV](#425-combatting-mev)
  - [5. Incentive Fairness Analysis](#5-incentive-fairness-analysis)
  - [6. Security Analysis](#6-security-analysis)
  - [7. Liveness Analysis](#7-liveness-analysis)
  - [8. License](#8-license)

## 1. Summary 
Each epoch in BSC consists of multiple slots, and a batch of validators take turns in a predefined order to obtain priority block-producing rights for each slot. This BEP proposes an adjustment to the allocation of priority block-producing rights: each validator receives priority block-producing rights for a predetermined number of consecutive slots per round.
## 2. Abstract 
BEP-341 describes a new allocation method for priority block-producing rights, which is logically concise and significantly enhances the system's transaction processing capacity, while remaining orthogonal to transaction processing optimization techniques within blocks.
## 3. Motivation 
The BSC ecosystem is active and continuously evolving, requiring ongoing improvements to the system's transaction processing capacity.
### 4.1 Scaling Principle 
Currently, each validator obtains priority block-producing rights for a single slot and is then rotated, with a fixed block interval of t. The transaction processing limit is t/2 for validating transactions from the previous block and t/2 for processing transactions in the new block.
<div align="center">

</div>
Assuming the number of transactions that can be processed is T, the average TPS is calculated as follows:
<div align="center">

</div>
After implementing this BEP, each validator can obtain priority block-producing rights for a continuous sequence of n slots in one round. The first block they process still allocates t/2 for validating transactions from the previous block and t/2 for processing transactions in the new block. However, subsequent blocks can skip the transaction validation process and focus on processing transactions in the new block. It is recommended that the processing time for the last block in their current round be only half of the time required to process transactions in the new block. This helps avoid the situation where the first block processed by the next validator is close to being an empty block.
<div align="center">

</div>
The TPS after continuous block production is as follows:
<div align="center">

</div>
Therefore, the TPS improvement rate is:
<div align="center">

</div>
The relationship between the number of continuous blocks produced and TPS improvement is 
<div align="center">

</div>
As shown in the graph, it is evident that continuous block production cannot double the TPS. There is no benefit when n<3, and increasing n beyond 5 does not proportionally increase the benefits. Therefore, a value within the range [3, 5] is suitable for n; when n=4, r=50%.

#### 4.2.1 Priority Allocation 
Each epoch predefines a set of validators, with a total of validatorN validators, and each validator within the set has a unique index ranging from [0, validatorN). If the current block height is blockN, then the validators with the following indices obtain priority block-producing rights.
<div align="center">

</div>

#### 4.2.2 Validator Set Switch 
Each epoch will choose a new validator set, assuming an epoch contains epochSlots slots. The validator set switch occurs only when the block height reaches Bswitch to prevent epoch block forging. The calculation of Bswitch is as follows:

<div align="center">

</div>

#### 4.2.3 Block Avoidance 
To prevent fewer than 1/2 of the nodes from controlling the entire network, block producers are required to produce fewer than n blocks within the previous ((validatorN/2+1)\*n-1) historical blocks.

#### 4.2.4 Governable Number of Consecutive Blocks 
When n=1, it is equivalent to disabling the feature of consecutive block production, while significant optimization is observed when n belongs to the range [3,5]. Currently, the range for the value of n is set to [1,9] but except 2.

The initial value of n is 1, and changing its value requires the BSC governance process.

#### 4.2.5 Combatting MEV
As the consecutive period in which a single validator gains priority in block production extends, it may facilitate MEV extraction, potentially leading validators to include more transactions in the later blocks they consecutively produce. To constrain validators to promptly package transactions, within a validator's consecutive priority over n blocks, the transaction fees' split to the SystemRewardContract will increase linearly with block number, capped at the value denoted as systemRewardAntiMEVRatio.
Respectively, the split ratio remains at systemRewardBaseRatio when continuous block production is disabled. Once continuous block production is enabled (i.e., when n > 1), the systemRewardRatio is calculated as:
<div align="center">

</div>
as shown in the following picture:
<div align="center">

</div>

The initial value of systemRewardAntiMEVRatio is 0, and changing its value also requires the BSC governance process.

## 5. Incentive Fairness Analysis
Within a single epoch, tail validators have fewer block-producing opportunities, but the allocation of priority rights is unbiased and cannot be manipulated. Therefore, from a statistical perspective, it is fair.

If a validator tries to extract more MEV by placing more transactions in later blocks they produce, it will also increase user transaction confirmation times. To curb this, the systemRewardAntiMEVRatio can be raised through the governance process, which will increase the proportion of transaction fees allocated to the SystemRewardContract. These bonuses will be distributed as Fast Finality voting rewards to validators, keeping their overall benefits unchanged. However, during high traffic and transaction backlog, high-performance validators usually handle more transactions. These changes will reduce the income advantage of having high performance.

## 6. Security Analysis
This BEP relies on BSC's Fast Finality feature. If Fast Finality fails, it may result in the following issues:
1. Nodes intentionally hide mined blocks, potentially leading to longer short-term reorganizations. 
2. The probability of finality for transactions increases, requiring waiting for 2/3\*validatorN\*n+ blocks. 

## 7. Liveness Analysis
The liveness of the chain remains unchanged, meaning it is required to ensure that at least (validatorN/2+1) validators are active. The proof is as follows:

If (validatorN/2+1) validators are active and none are allowed to produce blocks at a certain moment, each validator must have produced at least n blocks in the past ((validatorN/2+1)\*n-1) blocks. Thus, collectively, at least (validatorN/2+1)\*n blocks must have been produced, which is impossible. Therefore, at any moment, at least one of the (validatorN/2+1) validators must be allowed to produce blocks.

## 8. License 


NathanBSC:
https://github.com/bnb-chain/BEPs/pull/301

Magkoohp:
0x4ea83aebbd3d2c29cfe79737aa35b4302f26bcb6

Fdutoit96 (approved):

BEPs/BEP-341.md
```diff
@@ -0,0 +1,113 @@
...
+<img src=./assets/bep-341/4.2.2.png width=48% />
+</div>
+
+#### 4.2.3 Block Avoidance
+To prevent fewer than 1/2 of the nodes from controlling the entire network, block producers are required to produce fewer than n blocks within the previous validatorN/2\*n historical blocks.
+
+#### 4.2.4 Governable Number of Consecutive Blocks
+When n=1, it is equivalent to disabling the feature of consecutive block production, while significant optimization is observed when n belongs to the range [3,5]. Currently, the range for the value of n is set to [1,9] but except 2.
+
+#### 4.2.5 Combatting MEV
```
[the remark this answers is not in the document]
NathanBSC: > 

both are ok to me.
image

BEPs/BEP-341.md
```diff
@@ -0,0 +1,113 @@
...
+<img src=./assets/bep-341/4.2.5-2.png width=75% />
+</div>
+In addition, systemRewardAntiMEVRatio will be governable.
+
+## 5. Incentive Fairness Analysis
+Within a single epoch, tail validators have fewer block-producing opportunities, but the allocation of priority rights is unbiased and cannot be manipulated. Therefore, from a statistical perspective, it is fair.
+## 6. Security Analysis
+This BEP relies on BSC's Fast Finality feature. If Fast Finality fails, it may result in the following issues:
+1. Nodes intentionally hide mined blocks, potentially leading to longer short-term reorganizations. 
+2. The probability of finality for transactions increases, requiring waiting for 2/3\*validatorN\*n + blocks.
```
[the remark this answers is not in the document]
NathanBSC: here means 2/3*validatorN*n or more
I will discard the space before `+`

BEPs/BEP-341.md
```diff
@@ -0,0 +1,113 @@
...
+</div>
+The relationship between the number of continuous blocks produced and TPS improvement is 
+<div align="center">
+<img src=./assets/bep-341/4.1-6.png width=60% />
+</div>
+As shown in the graph, it is evident that continuous block production cannot double the TPS. There is no benefit when n<3, and increasing n beyond 5 does not proportionally increase the benefits. Therefore, a value within the range [3, 5] is suitable for n; when n=4, r=50%.
+
+### 4.2 Implementation Specification 
+#### 4.2.1 Priority Allocation 
+Each epoch predefines a set of validators, with a total of validatorN validators, and each validator within the set has a unique index ranging from [0, validatorN). If the current block height is blockN, then the validators with the following indices obtain priority block-producing rights.
```
emailtovamos, line 71: Could we elaborate a bit more here with an example? 
Say if blockN is 100, n = 4, validatorN is 10, which indices obtain priority? Also could you add brackets to make the formulas easier to understand?

buddh0: > Could we elaborate a bit more here with an example? Say if blockN is 100, n = 4, validatorN is 10, which indices obtain priority? Also could you add brackets to make the formulas easier to understand?

100/4%10 -->5
so the index is 5

buddh0:
### Question:
Why change the block history length to ((validatorN/2+1)*n-1) from validatorN/2*n?

### Answer:
A consensus algorithm's primary goal is to ensure liveness and consistency.

Based on liveness analysis, both values ensure the chain remains live when (validatorN/2+1) nodes are functioning correctly. However, using ((validatorN/2+1)*n-1) slightly improves consistency. For example, consider a clone attack with 21 nodes, one of which is malicious, and the remaining 20 are split into two subnets of 10 nodes each, with each node producing n consecutive blocks. According to liveness requirements, both subnets will eventually stop and lose liveness.
- If the block history length is validatorN/2*n, the malicious node can produce one block in each subnet, allowing the chain to continue.
- If the block history length is ((validatorN/2+1)*n-1), the malicious node must produce n blocks in each subnet for the chain to continue.
Thus, ((validatorN/2+1)*n-1) imposes stricter requirements on Byzantine nodes, enhancing consistency without compromising liveness.

Additionally, for backward compatibility considerations, before this BEP takes effect, with n==1:
- validatorN/2*n becomes validatorN/2
- ((validatorN/2+1)*n-1) also becomes validatorN/2
Therefore, both values are backward compatible.

NathanBSC:
### Question:
Why change the block history length to ((validatorN/2+1)\*n-1) from validatorN/2\*n?

### Answer:
A consensus algorithm's primary goal is to ensure liveness and consistency.

Based on liveness analysis, both values ensure the chain remains live when (validatorN/2+1) nodes are functioning correctly. However, using ((validatorN/2+1)\*n-1) slightly improves consistency. For example, consider a clone attack with 21 nodes, one of which is malicious, and the remaining 20 are split into two subnets of 10 nodes each, with each node producing n consecutive blocks. According to liveness requirements, both subnets will eventually stop and lose liveness.
- If the block history length is validatorN/2\*n, the malicious node can produce one block in each subnet, allowing the chain to continue.
- If the block history length is ((validatorN/2+1)\*n-1), the malicious node must produce n blocks in each subnet for the chain to continue.

Thus, ((validatorN/2+1)\*n-1) imposes stricter requirements on Byzantine nodes, enhancing consistency without compromising liveness.

Additionally, for backward compatibility considerations, before this BEP takes effect, with n==1:
- validatorN/2\*n becomes validatorN/2
- ((validatorN/2+1)\*n-1) also becomes validatorN/2

Therefore, both values are backward compatible.

huihzhao:
any APY impact for validators?

huihzhao (approved):

buddh0:
> 5. Incentive Fairness Analysis

pls refer `**5. Incentive Fairness Analysis**`

BEPs/BEP-341.md
zzzckck: may not need these .png files, insert latex like: $\(blockN/n)\mod{validatorN}$

huihzhao:
> Regarding 4.2.5 Combatting MEV, the BEP says "To constrain validators to promptly package transactions, within a validator's consecutive priority over n blocks, the transaction fees' split to the SystemRewardContract will increase linearly with block number... "
> 
> The problem with this approach is that it is not possible to force validators to share their MEV profit with delegators / SystemRewardContract.
> 
> So, if for example the validators make an extra MEV profit by producing consecutive blocks, they can keep this extra profit to themselves, bypassing SystemRewardContract. Therefore, the constraint made by systemRewardAntiMEVRatio is not really effective as it is limited only to the transaction fees, which is likely just a small fraction of the total profit that validators could potentially make by signing consecutive blocks.
> 
> The bottom line is that the extra profit from signing consecutive blocks is much higher than the validator profit from transaction fees and therefore validators can ignore the constraint made by systemRewardAntiMEVRatio and users will wait longer for transaction confirmations.
> 
> This change is definitely good for the validators, who will be able to aggressively extract more MEV by signing consequent blocks, but it is not clear how this change is good for the network, as it will result in lower confirmation time for the users.

As BNBChain has impelmented the PBS, the MEV revenue that through the standard builder API could be included in the gas fee. that is another consideration when wen discuss this topic. I assume, for normal users, they can also use a builder to be protected from attacks, and also of course for searchers.

NathanBSC:
rebase master and force push

zzzckck (approved):

Aces42020 (approved):

octavio12345300 (approved):
List

octavio12345300 (approved):
List

PR #418, opened 2024-08-05
BEP-418: increase initial backoff time to 2 second

zzzckck:


zzzckck:
https://forum.bnbchain.org/t/bep-418-increase-initial-backofftime-from-1-to-2-second/2925

PR #459, opened 2024-11-20
introducing Transaction Inspection Zone to block builder

TenArmor:


BEPs/BEP-467.md
```diff
@@ -0,0 +1,89 @@
...
+  - Inspection code is evm compatible and can be deployed on-chain for transparency, it is protocol specific, and developed by protocol developers in their projects, something like invariant tests.
+  - Protocols register their inspection code through open API, the registration let protocols decide the exact condition as to when to run their inspection code, and define the precise inspection point during the execution of a transaction.
+- **Response Action Engine**: Transactions identified as potentially malicious are flagged for rejection or additional review, this engine allows user to carry out predefined actions.
+  - Proactive Vulnerability Response: in the event of an attack occuring it is crucial to act as quickly as possible, to do so protocols can registers actions for different situation beforehand, such as pausing protocol, disabling transfering of funds, etc.
+  - Builtin standard actions: rejecting transaction for the next block, notifying relavant parties, etc.
+
+Once inspected, the cleared transactions are forwarded back to the block builder.
```
unclezoro, line 41: any design on validator side, like:
1. what if a validator does not integrate MEV?
2. or a validator do integrate with MEV but adopt local block eventually?

TenArmor: Validators can integrate with this system independently.

Current design does not necessarily involve any changes to the consensus or Geth codebase. We believe such a design brings ease of adoption at the builder level and maintains the overall stability of the blockchain.

gandonghb:
This proposal may have the following risks:

1.Over-reliance on predefined rules: the proposal emphasizes pre-registered actions, but the complexity and diversity of attacks may exceed the scope of these predefined rules. For example, some attacks may exploit unforeseen vulnerabilities in the protocol, causing the standard actions to fail. Additionally, malicious transactions might evade detection through obfuscation. As a result, we cannot fully ensure the absence of false negatives or false positives.

2.Potential impact on user experience and protocol credibility: If these protective mechanisms are triggered in non-attack scenarios (e.g., mistakenly pausing the protocol or rejecting transactions), it may result in user funds being frozen or transactions failing, which could harm the protocol’s reputation.

3.Impact on the efficiency of on-chain operations: Built-in standard actions, such as "rejecting transactions for the next block," need to closely integrate with block generation logic. This may lead to performance overhead or delays in transaction confirmation on-chain.

In summary, the current design in the proposal may introduce instability to the blockchain. To advance this proposal, we may need a more detailed description of the mechanism, such as:

1.Enhancing dynamic response capabilities based on real-time status. For instance, after detecting an attack, the response strategy adjusts dynamically depending on the type of attack rather than relying solely on predefined actions.

2.Consider the impact of false positives.

3.Considering the efficiency of on-chain operations and the overall stability of the protocol.

BEPs/BEP-467.md
8153764: > any design on validator side, like:
> 
> 1. what if a validator does not integrate MEV?
> 2. or a validator do integrate with MEV but adopt local block eventually?

octavio12345300 (approved):
listo

BEPs/BEP-459.md
8153764, line 1: Fish

8153764 (approved):

BEPs/BEP-459.md
8153764: Fish Finnish withdrow

BEPs/assets/bep-467/image-1.png
Mias88, line 1: BEPs/assets/bep-467/image-1.png

octavio12345300 (approved):
List

hilljessica8787-dot (approved):

PR #460, opened 2024-11-22
feat: bundle data API

irrun:
##### 4.4.1.3 Query Bundle Data

This API enables MEV data analysts to query bundle information from builders. 
It is particularly useful for MEV data providers who need to analyze market trends and create analytics dashboards. 
For instance, to accurately measure MEV market activity, providers can use this API to display metrics such as 
daily bundle counts and bundles per block in their analytics panels.

Request:

```json
{
  "jsonrpc": "2.0",
  "method": "mev_bundles",
  "params": [
    {
      "fromBlock": "0x1",
      "toBlock": "0x3"
    }
  ],
  "id": 1
}
```

Params:

- `fromBlock`: It is hex value of a block number.
- `toBlock`: Either the hex value of a block number OR block tags: "latest", "pending" or nil. Defaults to "latest". It
  should be no less than `fromBlock` and less than `fromBlock+100`.

Response:

* Return bundles whose first-received-height(`receivedBlock`)  is within the block range [`fromBlock`, `toBlock`].

```json
{
  "jsonrpc": "2.0",
  "result": [
    {
      "receivedBlock": "0x1",
      "bundles": [
        ["0x88...44b", "0x89..45b"], // tx hase list in the bundle
        ["0x90...46b"]
      ]
    },
    {
      "receivedBlock": "0x2",
      "bundles": [
        ["0x23...24b", "0x39..45b"],
        ["0x43...55a", "0x45..33a", "0x39..45b"]  //  it is possible that a same tx hash exists in different block
      ]
    }
  ],
  "id": 1
}
```

### DIFF
--- a/BEPs/BEP322.md
+++ b/BEPs/BEP322.md
@@ -11,26 +11,26 @@
 <!-- TOC -->
 
 - [BEP-322: Builder API Specification for BNB Smart Chain](#bep-322-builder-api-specification-for-bnb-smart-chain)
-    - [1. Summary](#1-summary)
-    - [2. Motivation](#2-motivation)
-    - [3. Background](#3-background)
-        - [3.1 BSC Trust Model](#31-bsc-trust-model)
-        - [3.2 BSC Consensus](#32-bsc-consensus)
-    - [4 Workflow](#4-workflow)
-        - [4.1 Definition](#41-definition)
-        - [4.2 One-Round Interaction (default)](#42-one-round-interaction-default)
-        - [4.3 Two-Round Interaction (optional)](#43-two-round-interaction-optional)
-        - [4.4 APIs](#44-apis)
-            - [4.4.1 Builder APIs](#441-builder-apis)
-                - [4.4.1.1 Retrieve Transactions](#4411-retrieve-transactions)
-                - [4.4.1.2 Issues Report](#4412-issues-report)
-            - [4.4.2 Validator APIs](#442-validator-apis)
-                - [4.4.2.1 Bid Block](#4421-bid-block)
-    - [5. Further Discussion](#5-further-discussion)
-        - [5.1 Implementation Consideration](#51-implementation-consideration)
-        - [5.2 Payment \& Economic Considerations](#52-payment--economic-considerations)
-        - [5.3 Builder Registration](#53-builder-registration)
-    - [6. License](#6-license)
+  - [1. Summary](#1-summary)
+  - [2. Motivation](#2-motivation)
+  - [3. Background](#3-background)
+    - [3.1 BSC Trust Model](#31-bsc-trust-model)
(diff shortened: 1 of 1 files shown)


BEPs/BEP322.md
```diff
@@ -288,6 +288,57 @@ Response:
...
+* The diff between fromBlock and toBlock should be less than 100.
+
+Response:
+
+```json
+{
+  "jsonrpc": "2.0",
+  "result": [
+    {
+      "blockNumber": "0x1",
```
unclezoro, line 321: does `blockNumber` means when receive this block?

BEPs/BEP322.md
```diff
@@ -288,6 +288,57 @@ Response:
...
+      "bundleCount": "0x2",
+      "bundles": [
+        {"txHashes": ["0x88...44b", "0x89..45b", "0x90...46b"]},
```
unclezoro, line 324: can each bundle have an expire block so that the index service can no long monitor the txs inside the bundle even though it may include in further block?

BEPs/BEP322.md
```diff
@@ -288,6 +288,57 @@ Response:
...
+Request:
+
+```json
+{
+  "jsonrpc": "2.0",
+  "method": "mev_bundles",
+  "params": [
+    {
+      "fromBlock": "0x1",
+      "toBlock": "0x2"
```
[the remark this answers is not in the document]
reneecok: I think it's absolutely web3 style, according to getLogs (https://docs.alchemy.com/reference/eth-getlogs)

BEPs/BEP322.md
irrun: refact the response to present bundle's lifetime

BEPs/BEP322.md
irrun: refact the response to present bundle's lifetime

BEPs/BEP322.md
```diff
@@ -288,6 +288,62 @@ Response:
...
+  should be no less than `fromBlock` and less than `fromBlock+100`.
+
+Response:
+
+```json
+{
+  "jsonrpc": "2.0",
+  "result": [
+    {
+      "fromBlock": "0x1",
```
unclezoro, line 326: will this introduce too many duplicated data for an indexer when it try to catch up?

Jackytan88 (approved):

BEPs/BEP322.md
irrun: Added a note to return only bundles whose height is within the block range when the bundle was first received, regardless of the bundle end time.

octavio12345300 (approved):

octavio12345300 (approved):

octavio12345300 (approved):

unclezoro (approved):

PR #530, opened 2025-02-28
BEP-530: Improved Block Producer Selection Mechanism

buddh0:
## Description
This BEP proposes a new block backup producer selection mechanism to improve the network stability.
This is the initial version of this BEP, pls refer the BEP directly for the latest content: https://github.com/buddh0/BEPs/blob/improve-block-producer-selection/BEPs/BEP-530.md


BEPs/BEP-530.md
```diff
@@ -0,0 +1,89 @@
...
+#### 4.1.2 Determining Turn Order
+At any given block height, one validator is designated as the "in-turn" validator with the highest priority to produce a block, with a block difficulty of 2. Additionally, multiple backup validators are assigned with a block difficulty of 1, and their blocks are broadcasted with a delay. The index of the in-turn validator at block height     `blockNumber` is determined as follows:
+```Go
+// turnLength specifies how many consecutive blocks a validator can produce
+inturnIndex := blockNumber / uint64(turnLength) % uint64(len(validators))
+```
+
+Backup validators are selected by reversing the order of the validators prior to the `inturnIndex`. The block producer sequence, including both in-turn and backup validators, is shown as follows:
+```Go
+// Reverse round
```
galaio, line 52: Why does here use reverse round? Is there any problem with positive order?

BEPs/BEP-530.md
```diff
@@ -0,0 +1,89 @@
...
+```Go
+// Reverse round	
+for order := 0; order < len(validators); order++ {
+			log.Debug("block producing", "order", order,
+				"validator index", (inturnIndex+len(validators)-order)%len(validators))
+}
+```
+
+#### 4.1.3 Final Producer Filtering
+To prevent chain splits, Parlia continuously records the recent block producers within a given length, known as `MinerHistory`. Once these validators reach a predefined block production threshold, they are no longer eligible to produce blocks at the current height. This BEP proposes using the latest finalized block to clean `MinerHistory`, removing producers who have produced blocks up to and including the latest finalized block.
```
galaio, line 60: Could you help to explain why it can prevent chain splits? Another question is does that undermine fairness? Maybe some validators will produce more blocks?

BEPs/BEP-530.md
```diff
@@ -0,0 +1,89 @@
...
+## 2. Abstract
+This BEP introduces a new algorithm for selecting block producers, offering improvements in four key areas: security, fairness, robustness, and simplicity.
+
+## 3. Motivation
+The motivation is to increase the ratio of blocks produced by the "in-turn" validator to enhance network stability. Additionally, it aims to improve the predictability of backup validators at each block height to support optimized transaction and block propagation.
+
+## 4. Specification
+### 4.1 Algorithm Overview
+#### 4.1.1 Validator Set Retrieval
+Each epoch involves a change in the validator set. After retrieving the new set from the contract, we use `epochNumber (blockNumber / EpochLength)` as the seed to randomly shuffle the validator set. Validators are assigned an index within the range `[0, len(validators))`, and this index remains unchanged for the duration of the epoch.
```
zzzckck, line 41:
```suggestion
Each epoch involves a change in the validator set. After retrieving the new set from the contract, we use `epochNumber(blockNumber/EpochLength)` as the seed to randomly shuffle the validator set. Validators are assigned an index within the range `[0, len(validators))`, and this index remains unchanged for the duration of the epoch.
```

BEPs/BEP-530.md
```diff
@@ -0,0 +1,89 @@
...
+This BEP proposes an improved block producer selection mechanism.
+
+## 2. Abstract
+This BEP introduces a new algorithm for selecting block producers, offering improvements in four key areas: security, fairness, robustness, and simplicity.
+
+## 3. Motivation
+The motivation is to increase the ratio of blocks produced by the "in-turn" validator to enhance network stability. Additionally, it aims to improve the predictability of backup validators at each block height to support optimized transaction and block propagation.
+
+## 4. Specification
+### 4.1 Algorithm Overview
```
zzzckck, line 39: maybe we can remove `4.1 Algorithm Overview`, not suggest overview in BEP specification.

BEPs/BEP-530.md
```diff
@@ -0,0 +1,89 @@
...
+
+When Fast Finality is not working properly, the new algorithm ensures that validators about to be in-turn are not prioritized as backup validators at the current block height, further reducing the chance of them being excluded by `MinerHistory` during their in-turn block production.
+
+In summary, this algorithm increases the likelihood of blocks being produced by the in-turn validator.
+
+#### 4.2.4 Simplicity Analysis
+The logic for selecting the in-turn validator remains unchanged, but the logic for selecting backup validators has been simplified. Particularly when Fast Finality is functioning properly, the sequence of block producers can be roughly viewed as validators being selected in reverse order from the `inturnIndex`.
+
+## 5. Backwards Compatibility
+A hard fork is required to implement this mechanism.
```
zzzckck, line 86: `Backwards Compatibility` is to show the impact to ecosystem.
I think validators will be impacted, as the block producer selection mechanism will be changed, it tries to randomize the block producer. Validator collusion can be mitigated, the opportunity to produce block would be faire in long term.

BEPs/BEP-530.md
zzzckck, line 41: I think the seed could be too simple and predicable.

zzzckck: accepted, as there is no other better choice, as blockHash could be manipulated, while epochNumber is fixed.

zzzckck (changes requested):
at discussion phase, not ready to merge

PR #532, opened 2025-03-03
update BEP-520: add two backward compatibility

buddh0:
add description for two cases about `Backward Compatibility`
1. Timing Based on Block Numbers
2. Indexing Based on `block.timestamp`

### DIFF
--- a/BEPs/BEP-520.md
+++ b/BEPs/BEP-520.md
@@ -82,7 +82,7 @@ This behavior needs to be penalized. In a continuous block production scenario, 
 The detection condition is as follows:
 ```Go
 // blockInterval     = time.Duration(1500)*time.Millisecond
-func (p *Parlia) isBlockDelayMinedOnPurpose(header, parent *types.Header) {
+func (p *Parlia) isBlockDelayMinedOnPurpose(header, parent *types.Header) bool {
 	return header.Difficulty == diffInTurn && parent.Difficulty == diffInTurn &&
 		header.Coinbase == parent.Coinbase &&
 		parent.MilliTimestamp()+blockInterval < header.MilliTimestamp()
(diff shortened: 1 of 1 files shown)


BEPs/BEP-520.md
```diff
@@ -152,7 +154,13 @@ The [Quarterly Auto-Burn](https://www.bnbburn.info/) mechanism also requires adj
...
 where T represents the number of seconds in the corresponding quarter.
 
-### 6.4 Block Finality
+### 6.4 Timing Based on Block Numbers
+The reduction in block intervals affects logic that relies on block numbers for timing, whether in independent services or within smart contracts. A simple solution is to adjust the corresponding block count or avoid using block numbers for timing altogether.
+
+### 6.5 Indexing Based on `block.timestamp`
+With the block interval reduced to 0.75 seconds, `block.timestamp`, which has second-level precision, may be the same for consecutive blocks. Therefore, using `block.timestamp` as an index key or for similar purposes requires adjustment. A common solution is to use the block hash instead.
+
+### 6.6 Block Finality
```
zzzckck, line 163:
```suggestion
### 6.4.DApp Developers And Users
**Timing Based on Block Numbers**
The reduction in block intervals affects logic that relies on block numbers for timing, whether in independent services or within smart contracts. A simple solution is to adjust the corresponding block count or avoid using block numbers for timing altogether.
**Indexing Based on `block.timestamp`**
With the block interval reduced to 0.75 seconds, `block.timestamp`, which has second-level precision, may be the same for consecutive blocks. Therefore, using `block.timestamp` as an index key or for similar purposes requires adjustment. A common solution is to use the block hash instead.

### 6.6 Block Finality
```

zzzckck: The two can be combinaed

zzzckck (approved):

PR #559, opened 2025-04-15
BEP-559: Encrypted Mempool for BNB Smart Chain

Smokyish:
Proposal for an Encrypted Mempool for Binance Smart Chain.

### DIFF
--- a/BEPs/BEP-547.md
+++ b/BEPs/BEP-547.md
@@ -1,0 +1,416 @@
+<pre>
+  BEP: 547
+  Title: Encrypted Mempool for Binance Smart Chain
+  Status: Proposed
+  Type: Standards & Improvements
+  Created: 2025-04-15
+</pre>
+
+# Summary
+
+**Summary/TL;DR**
+
+This proposal outlines an initiative to implement a distributed (threshold) encrypted mempool for the Binance Chain, aimed at thwarting front-running and adding a robust layer of real-time censorship resistance. Motivated by the ethical vision of the BNB Good Will Alliance---which seeks to foster best practices and safeguard user interests---this proposal suggests that validators on Binance Chain can take a practical step by committing to include encrypted transactions in block proposals. These encrypted transactions are ordered and committed before decryption, ensuring that transaction details remain hidden until after the block is finalized, effectively eliminating the opportunity for front-running, as has already been demonstrated on the Gnosis Chain.
+
+At a high level, the system allows users to submit transactions in an encrypted format while validators process them without decryption. Once the transaction order is locked in, the transactions are decrypted and executed using a distributed threshold encryption/decryption mechanism. The threshold decryption can be performed either by the Binance validators themselves or by a dedicated Shutter Keyper set, which is selected by a purpose-built DAO (Shutter DAO 0x36). This ensures that no single entity can control or manipulate the transaction order to their advantage or insert transactions with knowledge that they gained through unfair advantage.
+
+On an implementation level, validators that will use the upcoming Erigon v3 client on Binance Chain are going to have the option to register for processing encrypted transactions. This process integrates this feature without disrupting the existing workflow. Erigon is actively developing a second client implementation for this specific purpose so that a version of Erigon v3 will be capable of processing encrypted transactions. The new release and feature will soon be available for BSC. 
+
+Importantly, this mechanism is entirely opt-in: regular transactions will continue to be supported while users and RPCs gain the added choice of utilizing enhanced privacy features.
+
+This proposal is a joint effort by brainbot GmbH, Erigon, and NodeReal, reflecting our collective commitment to advancing the Binance ecosystem and supporting its validator community. While we are excited about the potential of an encrypted mempool to enhance transaction fairness, improve censorship resistance, and deter malicious MEV activities, this document does not yet commit to specific assignments, milestones or resources. Detailed implementation tasks and deliverables will be defined in subsequent phases, ensuring that the transition is smooth and leverages existing technology already proven on networks like the Gnosis Chain.
+
+## Motivation
+
+### Community Trust and Ethical Standards
+
+In recent years, challenges like front-running and sandwich attacks have eroded trust in blockchain transaction ordering. The BNB Good Will Alliance was established to address such challenges by driving off-chain best practices and ethical standards for the benefit of all ecosystem participants. This proposal is a natural extension of those efforts, ensuring that transaction ordering is transparent, secure, and fair.
+
+### Mitigating Malicious MEV and Front-Running
+
+Front-running exploits the order of transactions to yield undue profit, destabilizing market fairness. By committing to an encrypted mempool, validators secure transaction ordering by:
+
+- **Pre-Decryption Ordering**: Locking in the order of transactions without exposure to plaintext data.
+
+- **Decentralized Decryption Authority**: Requiring a threshold amount of parties to collaborate to perform decryption ensures that no single party can control or manipulate the transaction order for their own benefit or insert transactions using information obtained through unfair means.
+
+- **Proven Technology**: Deployments on chains like Gnosis have showcased the effectiveness of such approaches in eliminating front-running.
+
+### Ecosystem Growth and Long-Term Vision
+
+Implementing this proposal will:
+
+- **Enhance User Confidence**: A secure, censorship-resistant transaction order process will attract more users and developers.
+
+- **Pave the Way for Advanced Integrations**: By aligning with the existing Proposer-Builder Separation supply chain standards and eventual Ethereum L1 integration, the encrypted mempool supports broader cross-chain innovations.
+
+- **Mitigate Regulatory Risks**: By managing MEV responsibly, Binance Chain validators might be able to avoid regulatory classifications that might otherwise label validators or sequencers as financial intermediaries and require extensive monitoring/reporting of market abuse. See the latest ESMA regulation in this regard. The analogy here is that a provider of an end-to-end encrypted messaging provider can also not be held liable for malicious activity happening on its encrypted technical infrastructure.
+
+### Real-Time Censorship Resistance and Builder Decentralization
+
+Increasing centralization of block builders is a growing concern on blockchain platforms, notably on Ethereum, where centralized builder models can lead to real-time censorship and unfair transaction ordering. Although Binance Chain has not yet experienced severe centralization issues, this proposal proactively addresses potential future risks by:
+
+- **Enhancing Censorship Resistance**: Encrypted transactions remain concealed until inclusion, preventing centralized actors from selectively censoring or reordering transactions based on their contents.
+
+- **Reducing Builder Centralization**: By decentralizing both transaction ordering (which is already the case on BSC) and decryption through threshold encryption/decryption, the power to curate transactions is distributed among a broad set of validators or Keypers. This discourages the formation of centralized builder pools and mitigates any risk of undue influence over transaction inclusion.
+
+- **Fostering Fairness**: A distributed and transparent ordering mechanism ensures that no single entity can dominate transaction sequencing, paving the way for a more equitable ecosystem.
+
+## Specification
+
+### Client Integration and Upgrade Path
+
+Validators on the Binance Chain can opt-in to encrypted transaction processing by upgrading to the latest Erigon v3 client. Key aspects include:
+
+- **Optional Registration**: Validators may register to support and include encrypted transactions while also continuing to process regular transactions.
+
+- **Seamless Transition**: The protocol remains backward-compatible, enabling a gradual and safe transition process.
+
+- **Multi-Client Support**: In addition to Erigon v3, similar support is anticipated from other clients like Nethermind, ensuring broad validator participation. Nethermind was the first client to support the encrypted mempool on the Gnosis Chain, so it is highly likely that it'll be trivial for them to support an implementation on BSC as well.
+
+### Threshold Encryption Workflow
+
+The heart of this proposal is the threshold encryption mechanism, which unfolds as follows:
+
+1. **Transaction Submission:**
+
+    Users encrypt transactions using a public key tied to the threshold committee, ensuring all submissions remain indistinguishable from one another.
+
+2. **Validator Commitment:**
+
+    Validators commit to including encrypted transactions in blocks while they are still encrypted. Each validator cryptographically commits to the transaction order.
+
+3. **Decentralized Decryption:**
+
+    Once a validator has committed to a bloc, a threshold decryption process is initiated. A committee---either composed of Binance validators or a dedicated Shutter Keyper set (managed by Shutter DAO 0x36)---collaborates to decrypt the transactions only when the required threshold is reached.
+
+This workflow decentralizes trust, ensuring that even if a subset of participants is compromised, the encrypted contents remain secure.
+
+### Integration with PBS and Cross-Chain Protocols
+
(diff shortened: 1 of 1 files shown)


zzzckck:
hi @Smokyish , thank you for your proposal, but I wonder if BSC really need the EncryptedMempool.
There is a discussion thread: https://forum.bnbchain.org/t/call-for-feedbacks-bsc-short-block-interval-validator-dedicated-network-and-direct-mempool/3348, lots of reply are against the PrivateTxPool concept.

pepae:
> hi @Smokyish , thank you for your proposal, but I wonder if BSC really need the EncryptedMempool. There is a discussion thread: https://forum.bnbchain.org/t/call-for-feedbacks-bsc-short-block-interval-validator-dedicated-network-and-direct-mempool/3348, lots of reply are against the PrivateTxPool concept.

Hey! I think those are two different things and I agree, that a private mempool is not a good choice:

**Private mempool:** Exclusive, non-public mempool which someone needs to control. Transactions are **simply hidden for everyone except for the validator or centralized mempool operator.** This is basically how most rollups work. It requires massive trust in whoever operates that mempool not to front-run/censor and is generally considered to be a force for centralization.

**Threshold encrypted mempool:** Transactions are encrypted in a decentralized fashion (using threshold encryption) until the order and inclusion is finalized/committed to by validators. So **no one can see the transactions** before they're fixed to go into the block, making it very hard to front-run in general.
This doesn't prevent the back-running/general arbitrage type MEV, which can still be extracted/distributed and lead to higher market efficiency.

More on this: https://ethresear.ch/t/the-road-towards-a-distributed-encrypted-mempool-on-ethereum/21717

Happy to explain more/answer any questions!

im-hazyn:
@zzzckck 

BEP-525 is for Private Mempool
BEP-559 is for Encrypted Mempool

Two different things.

Private mempool is a terrible idea (as you can see in the replies), and I really hope you guys don't go in that direction and drop BEP-525 completely.
Encrypted mempool, in the other hand, can be beneficial if implemented correctly. The links shared above are good starting points.

zzzckck:
> > hi @Smokyish , thank you for your proposal, but I wonder if BSC really need the EncryptedMempool. There is a discussion thread: https://forum.bnbchain.org/t/call-for-feedbacks-bsc-short-block-interval-validator-dedicated-network-and-direct-mempool/3348, lots of reply are against the PrivateTxPool concept.
> 
> Hey! I think those are two different things and I agree, that a private mempool is not a good choice:
> 
> **Private mempool:** Exclusive, non-public mempool which someone needs to control. Transactions are **simply hidden for everyone except for the validator or centralized mempool operator.** This is basically how most rollups work. It requires massive trust in whoever operates that mempool not to front-run/censor and is generally considered to be a force for centralization.
> 
> **Threshold encrypted mempool:** Transactions are encrypted in a decentralized fashion (using threshold encryption) until the order and inclusion is finalized/committed to by validators. So **no one can see the transactions** before they're fixed to go into the block, making it very hard to front-run in general. This doesn't prevent the back-running/general arbitrage type MEV, which can still be extracted/distributed and lead to higher market efficiency.
> 
> More on this: https://ethresear.ch/t/the-road-towards-a-distributed-encrypted-mempool-on-ethereum/21717
> 
> Happy to explain more/answer any questions!


Hi @pepae , I have read the post, it is a very good idea, but:
- 1.most of the front run can be avoided by protected-RPC service, it is much simple
- 2.concerns of "EncryptedMempool": its complexity and performance downgrade

pepae:
> > > hi @Smokyish , thank you for your proposal, but I wonder if BSC really need the EncryptedMempool. There is a discussion thread: https://forum.bnbchain.org/t/call-for-feedbacks-bsc-short-block-interval-validator-dedicated-network-and-direct-mempool/3348, lots of reply are against the PrivateTxPool concept.
> > 
> > 
> > Hey! I think those are two different things and I agree, that a private mempool is not a good choice:
> > **Private mempool:** Exclusive, non-public mempool which someone needs to control. Transactions are **simply hidden for everyone except for the validator or centralized mempool operator.** This is basically how most rollups work. It requires massive trust in whoever operates that mempool not to front-run/censor and is generally considered to be a force for centralization.
> > **Threshold encrypted mempool:** Transactions are encrypted in a decentralized fashion (using threshold encryption) until the order and inclusion is finalized/committed to by validators. So **no one can see the transactions** before they're fixed to go into the block, making it very hard to front-run in general. This doesn't prevent the back-running/general arbitrage type MEV, which can still be extracted/distributed and lead to higher market efficiency.
> > More on this: https://ethresear.ch/t/the-road-towards-a-distributed-encrypted-mempool-on-ethereum/21717
> > Happy to explain more/answer any questions!
> 
> Hi @pepae , I have read the post, it is a very good idea, but:
> 
> * 1.most of the front run can be avoided by protected-RPC service, it is much simple
> * 2.concerns of "EncryptedMempool": its complexity and performance downgrade

Good points! Let me address:

1. While being a good quick fix for some of the MEV related issues, conceptually, a protected RPC comes with similar centralization vectors and trust assumptions as a private mempool. In this case you'll need to trust the relay and the group of validators not to front-run. A lot of the comments in your mentioned thread are arguing against centralized control of the mempool and I'm afraid protected RPCs have that same issue.

More detail: protected RPCs still hinge on an off‑chain relay, and that relay is a single choke‑point you have to trust completely. It sees your raw tx before inclusion and could leak or reorder it.

So while protected RPCs hide you from public frontrunners, they hand the same power to an even smaller cartel of relay operators (probably 1 or 2). Until we eliminate the relay entirely with in‑protocol threshold encryption or enshrined PBS, protected‑RPCs carry the same centralization risk as a private mempool (plus one extra hop of trust).

2. There is some implementation complexity, but that work is already done on Gnosis Chain and Nethermind and Erigon clients support this already (Gnosis Chain is a "99%" Ethereum-like chain in terms of that client software, similar to BNB).

On performance: 2 separate concerns 1) throughput and 2) latency: if correctly implemented, there should be close to no impact to throughput (because the encryption/decryption scales well and is extremely light).
As for latency, if there's full buy in from the validators, the decryption adds only a small amount of latency (on the order of a couple 100 miliseconds, depending on network topology of the threshold committee nodes). Would think this is acceptable for a decentralized chain which has e.g. 4 second block time.

In any case, the having an encrypted mempool does not eliminate the option to send through non-encrypted transactions. It's just an option that is offered to users/wallets/dapps. Many transactions shouldn't go through encryption (e.g. if it's a simple token transfer which cannot be front-run).

pepae:
> We believe encrypted mempool introduces additional costs, primarily in the following aspects:
> 
> Additional system complexity and computational overhead from transaction encryption:
> 
> * A key question is who performs the transaction encryption—wallets, RPC providers, or others? Currently, none of these options seem ideal.Decryption also burdens validators, especially considering BSC's push toward millisecond-level block times.
> * Reduced revenue for validators and users : Encrypted mempool negatively impacts the efficiency and profitability of backrun strategies and other MEV strategies, ultimately leading to lower validator earnings.
> 
> On the other hand, encrypted mempool offers two advantages:
> 
> * Frontrunning prevention
> * Stronger censorship resistance
> 
> To achieve these goals, alternative solutions like execution ticket / pre-confirmation mechanisms might be more effective.

Great questions, let me try to answer:

1) so on who performs the transaction encryption, the nice thing is, both of your options can work alongside each other, plus we can also encrypt within the dapp frontend.

On Gnosis Chain, there's an encrypting RPC: https://github.com/shutter-network/encrypting-rpc-server (you can try this out for yourself here: https://blog.shutter.network/shielded-trading/#get-protected-on-gnosis-chain )

And here's an example of encryption within the dapp frontend: https://shutter.gnosischain.com/

2) As for reduced revenue for validators: This is true in the sense that it reduces the front-running types of MEV. But that's exactly the value that's currently lost by users. So we think this is a good thing (users will get better prices). 

If implemented correctly, back-running and general arbitrage should not be affected. Longer term, we'd expect that if we can build a fair marketplace that is a more level playing field and thus open to more mainstream users, we think overall (benign/positive) MEV rewards for validators then might even increase.

3) In regards to pre-confirmation/execution ticket, this can be very complementary with an encrypted mempool. Please check out our roadmap for how to integrate Shutter with proposer commitments (which are the basis of pre-confs): https://ethresear.ch/t/the-road-towards-a-distributed-encrypted-mempool-on-ethereum/21717
Tengentially related: we also think it's very complementary with ePBS and we'll publish a post on this very soon.

BlockRazorinc:
> > We believe encrypted mempool introduces additional costs, primarily in the following aspects:
> > Additional system complexity and computational overhead from transaction encryption:
> > 
> > * A key question is who performs the transaction encryption—wallets, RPC providers, or others? Currently, none of these options seem ideal.Decryption also burdens validators, especially considering BSC's push toward millisecond-level block times.
> > * Reduced revenue for validators and users : Encrypted mempool negatively impacts the efficiency and profitability of backrun strategies and other MEV strategies, ultimately leading to lower validator earnings.
> > 
> > On the other hand, encrypted mempool offers two advantages:
> > 
> > * Frontrunning prevention
> > * Stronger censorship resistance
> > 
> > To achieve these goals, alternative solutions like execution ticket / pre-confirmation mechanisms might be more effective.
> 
> Great questions, let me try to answer:
> 
> 1. so on who performs the transaction encryption, the nice thing is, both of your options can work alongside each other, plus we can also encrypt within the dapp frontend.
>    On Gnosis Chain, there's an encrypting RPC: https://github.com/shutter-network/encrypting-rpc-server (you can try this out for yourself here: https://blog.shutter.network/shielded-trading/#get-protected-on-gnosis-chain )
>    And here's an example of encryption within the dapp frontend: https://shutter.gnosischain.com/
>    Also, keep in mind, using the encrypted mempool is optional for users/wallets/dapps, so this won't impede any existing applications or applications which aren't interested in benefiting from transaction encryption.
> 2. As for reduced revenue for validators: This is true in the sense that it reduces the front-running types of MEV. But that's exactly the value that's currently lost by users. So we think this is a good thing (users will get better prices).
>    If implemented correctly, back-running and general arbitrage should not be affected. Longer term, we'd expect that if we can build a fair marketplace that is a more level playing field and thus open to more mainstream users, we think overall (benign/positive) MEV rewards for validators then might even increase.
> 3. In regards to pre-confirmation/execution ticket, this can be very complementary with an encrypted mempool. Please check out our roadmap for how to integrate Shutter with proposer commitments (which are the basis of pre-confs): https://ethresear.ch/t/the-road-towards-a-distributed-encrypted-mempool-on-ethereum/21717
>    Tengentially related: we also think it's very complementary with ePBS and we'll publish a post on this very soon.

I think encrypting RPC has same problem as protected RPCs, and we believe we need more comments from wallet and front end Dapp provider for this case.

Could you explain why back-running and arbitrage will not be affected and at what user case the overall MEV will increase?

pepae:
> I think encrypting RPC has same problem as protected RPCs, and we believe we need more comments from wallet and front end Dapp provider for this case.
> 


Agreed that an encrypting RPC in this context isn't the ideal end goal and introduces another trust assumption. For Gnosis Chain, it's seen as a stop gap solution towards the end goal, which is encryption by the users themselves, either in the wallet or dapp.
But in general, a validator level encrypted mempool with multiple avenues to encrypt, including the RPC is very different from a protected mempool which hinges fully on the RPC alone.

Btw, we think that the UX paradigm might be changing anyway, towards a more agentic UX. This could be a good opportunity also to implement encryption on the user side. Example: https://x.com/bezzenberger/status/1895137585414160393

Also note that today already, anyone can run their own RPC, so users can already today be end-to-end protected, if they wish to do so.

> Could you explain why back-running and arbitrage will not be affected and at what user case the overall MEV will increase?

Short answer is that the validator can still insert transactions immediately after the shielded transactions, just not before. So they can still back-run, and solutions like flashbots could be used to extract/distribute this type of back-running MEV.

We think overall MEV could increase over the long term, because we believe that by creating a fair marketplace that prioritizes accessibility and inclusion (i.e. making DEFI available not just to sophisticated traders, who know about MEV, but to an even bigger, broader, more mainstream audience, which is currently side-lined) we’ll draw in more users, which could actually contribute to an increase in overall transaction volume and thus validators’ back-running related MEV rewards.

48ClubIan:
go to opbnb is you want this

WTDuck2255:
I don't understand i touchtouchtouch everything i can

MRHUNG369 (commented):
ok

MRHUNG369:
**### _

Đề xuất BEP-559 (về cơ bản là BEP-547) áp dụng phương pháp mempool được mã hóa ngưỡng. Các giao dịch được gửi đến mempool sẽ được mã hóa, vì vậy trình xác thực, trình tạo hoặc bot MEV không thể xem nội dung ở front-run/sandwich. Khi đơn đặt hàng trong khối bị "khóa", giao dịch sẽ được giải mã và thực hiện. Do đó, các giao dịch "mùn thải MEV" (ví dụ: chạy trước, bánh sandwich) không thể bị gián đoạn, giúp mạng tránh tắc nghẽn do MEV txs.@>FALLONP /### _txs. ###**

MRHUNG369 (approved):
phê duyệt

MRHUNG369:
> > How will this proposal avoid BNB chain just being clogged up with spam MEV txs as described here? https://writings.flashbots.net/mev-and-the-limits-of-scaling
> 
> 
> 
> An encrypted mempool would remove malicious MEV like front-running and sandwich attacks. Arbitrage, liquidations and backrunning would still be possible, but it could alleviate some of the spam related to those.

Thank you for raising this important question.
While BEP-559 may not completely eliminate all forms of MEV, it provides a substantial improvement in limiting harmful activities like front-running, back-running, and MEV-driven spam attacks.

	By encrypting the mempool, this proposal effectively hides transaction content during the critical propagation window, making it significantly harder for MEV bots to exploit pending transactions. This helps to reduce network congestion, protect user intent, and preserve fairness in transaction ordering.

	Arbitrage and liquidation may still occur post-block inclusion, but those are structurally different from pre-execution attacks.

	Overall, I believe BEP-559 is a valuable step forward for the BNB ecosystem and should be supported.

pepae:
> How will this proposal avoid BNB chain just being clogged up with spam MEV txs as described here? https://writings.flashbots.net/mev-and-the-limits-of-scaling

Good question! Flashbot's proposal in the link you provided to auction off the backrunning slot would be a good way to achieve this, while maintaining front-running resistance. So this and the encrypted mempool should be complementary.

MRHUNG369 (approved):

davesmail1401-sudo (approved):

octavio12345300 (approved):
List

davesmail1401-sudo (approved):

MRHUNG369 (commented):
Ok

hailuu01041990-lab (approved):

caravello00:
Hey all, it looks like some sandwich attacks have returned. Could be a good time to revisit the encrypted mempool BEP as a potential longterm solution.

Sharing some screenshots of charts from the Dune dashboard below. But perhaps someone else has a better data source.
https://dune.com/hildobby/sandwiches?Blockchain=bnb
Screenshot 2025-12-23 at 12 39 28 PM

Screenshot 2025-12-23 at 12 39 33 PM

samueldreadfulwater3-hash (approved):

Mohammedsapry (approved):

Mohammedsapry (approved):

Mohammedsapry:
God

hilljessica8787-dot (approved):

PR #566, opened 2025-04-21
Update BEP-520.md

zzzckck:
Update the phase2 parameter of blob target and maximum. In consideration of L2 backward compatibility, keep the blob limit unchanged.


buddh0 (approved):

galaio (approved):

krud491:
GREAT

PR #637, opened 2025-11-19
BEP-619: change TurnLength to 8 from 16

allformless:
BEP-619: change TurnLength to 8 from 16


NathanBSC:
no need to include it in BEP

PR #657, opened 2026-01-22
BEP-657: Limit Blob Transaction Inclusion by Block Number

zlacfzy:
https://github.com/zlacfzy/BEPs/blob/BEP-X/BEPs/BEP-657.md


NathanBSC:
mark: then calculate blobbasefee based on blockN-10？

PR #668, opened 2026-03-03
BEP-668: BSC Pre-confirmation for Prop AMM

BlockRazorinc:


BlockRazorinc:
>Who is allowed to become a whitelisted Prop AMM?

Since Prop AMM is currently passively invoked by smart routers, and given that most smart routers have their own vetting mechanisms for Prop AMM, we propose that the admission of Prop AMM be jointly reviewed by the leading smart router providers on BSC to facilitate its integration.

>How to determine whether a transaction is a Prop AMM trade?

I believe some technical criteria can be used to identify Prop AMMs—for example, by checking if a transaction only updates the state of pre-declared Prop AMM addresses.

>What incentive would validators have to support this non-general protocol?

From a validator's perspective,the addition of Prop AMM quote update transactions creates a substantial new revenue stream. Drawing from experience on Solana, the daily transaction costs for price updates in a single Prop AMM pool can reach several hundred dollars; with nearly ten such pools in operation, these transactions serve as a meaningful supplement to validator income.

>Why should a chain make such major compromises specifically for the Prop AMM protocol? 
And if other protocols—like oracles/lending—have similar requirements in the future, what then?

We recognize that transaction ordering is a market problem. That said, the current Gas Price/Fee-based sorting mechanism is too restrictive and doesn't fully account for the diverse needs of the future DeFi ecosystem.

Consider OFA scenarios: if value-based ordering is the sole metric, transactions that prioritize validator bribes over user rebates will always win. Consequently, the system would favor transactions that extract more value from users as long as they pay a higher premium to the validator.

We view Prop AMMs as a vital enhancement to the DeFi ecosystem, offering superior pricing and deeper liquidity. However, the current transaction ordering logic on BSC fails to accommodate their essential requirement for low-cost, high-priority execution. This fundamental gap is the primary driver behind our proposal.

hilljessica8787-dot (approved):

PR #694, opened 2026-06-04
fix : improve  gas and time description

flywukong:
improve gas and time description of bep675


BEPs/BEP-675.md
```diff
@@ -89,7 +89,7 @@ Upon receiving BidBlocks, the validator tracks the current best by derived `GasF
 
 **1. Pre-seal header verification**
 
-Validate the builder-supplied header against locally derived consensus rules: block number, parent hash, coinbase, gas limit, base fee, difficulty must match deterministic values from the parent block; timestamp must pass the existing `blockTimeVerifyForRamanujanFork` range check. The derived `GasFee` must also be non-zero. Any failure causes immediate rejection.
+Validate the builder-supplied header against locally derived consensus rules: block number, parent hash, coinbase, gas limit, base fee, and difficulty must match deterministic values from the parent block; the header millisecond timestamp must be no earlier than `blockTimeVerifyForRamanujanFork` allows and no later than the value derived by `blockTimeForRamanujanFork`. The derived `GasFee` must also be non-zero. Any failure causes immediate rejection.
```
allformless: the header millisecond timestamp must be no earlier than `blockTimeVerifyForRamanujanFork` allows and no later than the value derived by `blockTimeForRamanujanFork`.

1. blockTimeVerifyForRamanujanFork --> blockTimeForRamanujanFork
2. blockTimeForRamanujanFork is not a fixed value, so no earlier and no later will lead to just use it

--> 
the header millisecond timestamp must be derived in the same way as the validator.

allformless (approved):

mjoody20 (commented):
Changs

PR #703, opened 2026-07-17
BEP-703: payment lane

joey0612:
Related:
- BEP: https://github.com/bnb-chain/BEPs/pull/703
- BSC client: https://github.com/bnb-chain/bsc/pull/3793
- Contract: https://github.com/bnb-chain/bsc-genesis-contract/pull/682


mikwespinoza-ux (approved):

hilljessica8787-dot (approved):

BEPs/BEP-703.md
```diff
@@ -0,0 +1,207 @@
...
+
+```
+signal(n−1) = generalGasUsed(n−1) / GasLimit(n−1)
+```
+
+- `signal >= EXPAND_TRIGGER_RATIO` — the chain is congested: the quota ratio increases by `EXPAND_STEP`, up to the effective maximum.
+- `signal < SHRINK_TRIGGER_RATIO` — the chain has slack: the quota ratio decreases by `SHRINK_STEP`, down to the effective minimum.
+- Otherwise — the hysteresis band: the quota is unchanged, absorbing fluctuation and preventing oscillation.
+
+Note that `generalGasUsed` includes payment gas consumed beyond the quota (that gas competed as general traffic), so the signal responds to both pressure sources: general demand squeezing available space, and payment demand overflowing its own floor.
```
will-2012: A bit different from the above: `generalGasUsed  = total gas consumed by general transactions in the block`.

BEPs/BEP-703.md
```diff
@@ -0,0 +1,207 @@
...
+- **Payment demand falls short of the quota** (`paymentGasUsed < paymentLaneSize`): the shortfall reduces the gas available to general transactions one-for-one. The unused reservation is deliberately left idle for that block rather than released — this is what makes the guarantee real ([§4](#4-rationale)).
+
+The rule constrains only gas totals. It says nothing about where transactions sit inside the block: payment and general transactions may be interleaved arbitrarily, and block producers keep full freedom over transaction selection and ordering within the accounting constraint. No new ordering rule, priority rule, or fee rule is introduced.
+
+### 3.3 Dynamic Quota Adjustment
+
+`paymentLaneSize` for block `n` is a deterministic function of chain history, computed by every node from block `n−1`; no new header field is required. The adjustment reads a single signal:
```
will-2012: Todo: think more. Restart / snap-sync maybe need to walk all the history to rebuild?

will-2012: reuse header field?? header extra??

BEPs/BEP-703.md
```diff
@@ -0,0 +1,207 @@
...
+
+```
+signal(n−1) = generalGasUsed(n−1) / GasLimit(n−1)
+```
+
+- `signal >= EXPAND_TRIGGER_RATIO` — the chain is congested: the quota ratio increases by `EXPAND_STEP`, up to the effective maximum.
```
will-2012: mark: quota ratio -> quota size, because the gas limit may be different, it is better to use the absolute value here.

hilljessica8787-dot (approved):

BEPs/BEP-703.md
stsminingempire-spec: Yea you might be right…… sorry

BEPs/BEP-703.md
```diff
@@ -0,0 +1,503 @@
...
+generalGasUsed   = the block's gas used less paymentGasUsed
+paymentLaneQuota = the gas this block reserves for payment transactions (computed per 3.4)
+```
+
+`generalGasUsed` is a residual, not a sum over general transactions, and the difference is not cosmetic. Parlia appends unsigned system transactions to the end of every block — validator reward, slashing, finality reward, and on a breathe block a validator-set update that has reached 12M gas on mainnet — and their gas is in the header's gas used. Defining the residual puts that gas on the general side without asking [§3.2](#32-transaction-classification) to classify a transaction no user submitted; one of those system transactions is a bare value transfer, so the classification route would instead turn on its destination happening to hold code. The obligation to include them also bounds the quota, in [§3.4.4](#344-bounds).
+
+The block MUST satisfy:
+
+```
+generalGasUsed + max(paymentGasUsed, paymentLaneQuota) <= GasLimit
```
NathanBSC: I think It's not compatible with https://eips.ethereum.org/EIPS/eip-7778.
for example , general tx refund 10% gas
so when generalGasUsed = 90%* GasLimit,
then the verification will pass, but no payment tx will be inserted.

maybe modify as following?:
```
paymentGasUsed + (GasLimit-GasUsed) >= paymentLaneQuota
```

BEPs/BEP-703.md
```diff
@@ -0,0 +1,503 @@
...
+
+With the common conditions, a bare transfer's intrinsic cost is exactly 21000 gas: `to` non-nil rules out contract creation, empty `data` rules out calldata cost, and the access-list and type restrictions rule out everything else chargeable before execution. Its other two tests are about substance rather than cost. A non-zero `value` is what makes the transaction a payment at all: a zero-value transfer moves nothing, so reserved gas would be protecting nothing. No code at `to` is what keeps the transfer from running contract code: an account that gains code earlier in the same block, by deployment or by an [EIP-7702](https://eips.ethereum.org/EIPS/eip-7702) authorisation, holds it by the time the transfer is classified and is general, as is one that already carried a delegation designator.
+
+```
+classify(tx, parentState, liveState):    # parentState: end of h-1.  liveState: just before tx runs.
+    if tx.to == nil:                                     return GENERAL    # common
+    if tx.type not in {0x00, 0x01, 0x02}:                return GENERAL    # common
+    if tx.accessList is not empty:                       return GENERAL    # common
```
NathanBSC: why force this?

BEPs/BEP-703.md
```diff
@@ -0,0 +1,503 @@
...
+
+    if tx.to in paymentContractList(parentState):        return PAYMENT    # listed, and stops here
+
+    if tx.data is empty                                                    # bare transfer
```
NathanBSC: why force this?

BEPs/BEP-703.md
NathanBSC, line 144: after https://eips.ethereum.org/EIPS/eip-7778.
these ‘xxxGasUsed’ should be count before refund

BEPs/BEP-703.md
will-2012: Remove this restriction to keep the rules simple. Many constraints were originally added to make trading as simple as possible, but in retrospect, they are unnecessary.

BEPs/BEP-703.md
will-2012: Remove this restriction to keep the rules simple. Many constraints were originally added to make trading as simple as possible, but in retrospect, they are unnecessary.

BEPs/BEP-703.md
will-2012: Got it, thanks.

BEPs/BEP-703.md
```diff
@@ -0,0 +1,287 @@
...
+
+```
+classify(tx, parentState, liveState):    # parentState: end of h-1.  liveState: before this tx applies.
+    if isSystemTransaction(tx):                          return GENERAL    # common
+    if tx.to == nil:                                     return GENERAL    # common
+    if tx.type not in {0x00, 0x01, 0x02}:                return GENERAL    # common
+
+    if tx.to in paymentContractList(parentState):        return PAYMENT    # listed, and stops here
+
+    if tx.value != 0                                                       # bare transfer
```
zlacfzy: Should the bare-transfer rule exclude precompiles? They have no bytecode, so adding 1 wei lets computation use the payment lane.

will-2012 (dismissed):

will-2012 (dismissed):

will-2012 (approved):

allformless:
approved

mikwespinoza-ux (approved):

PR #710, opened 2026-09-03
BAP-710: Unified PropAMM Relayer

asiawildboar:
https://github.com/asiawildboar/BEPs/blob/bap-propamm/BAPs/BAP-710.md

### DIFF
--- a/BAPs/BAP-710.md
+++ b/BAPs/BAP-710.md
@@ -1,0 +1,592 @@
+```
+BAP: 710
+Title: Unified PropAMM Relayer
+Status: Draft
+Type: Application
+Created: 2026-09-03
+Description: A builder-neutral execution layer for proprietary AMMs (PropAMMs) on BNB Chain, with a unified on-chain PropAMM standard. Makers stream quotes to one Relayer; takers keep using standard EVM JSON-RPC; the Relayer bundles matched quotes with taker transactions and broadcasts to every builder.
+```
+
+# BAP-710: Unified PropAMM Relayer
+
+- [BAP-710: Unified PropAMM Relayer]
+  - [1. Summary](#1-summary)
+  - [2. Abstract](#2-abstract)
+  - [3. Motivation](#3-motivation)
+    - [3.1 PropAMM Needs a Shared Execution Layer](#31-propamm-needs-a-shared-execution-layer)
+    - [3.2 Who Carries the Complexity Today](#32-who-carries-the-complexity-today)
+    - [3.3 Why One Oracle and One Pool Interface](#33-why-one-oracle-and-one-pool-interface)
+  - [4. Specification](#4-specification)
+    - [4.1 Roles and Architecture](#41-roles-and-architecture)
+    - [4.2 PropAMM Oracle: Priority Update Registry](#42-propamm-oracle-priority-update-registry)
+    - [4.3 PropAMM Pool Interface](#43-propamm-pool-interface)
+    - [4.4 Relayer Endpoint and Authentication](#44-relayer-endpoint-and-authentication)
+    - [4.5 Taker API](#45-taker-api)
+    - [4.6 Maker API](#46-maker-api)
+    - [4.7 Maker Protection](#47-maker-protection)
+    - [4.8 Analytics and Monitoring](#48-analytics-and-monitoring)
+    - [4.9 Open Protocol Scope](#49-open-protocol-scope)
+  - [5. Rationale](#5-rationale)
+    - [5.1 Design Decisions](#51-design-decisions)
+    - [5.2 Alternative Approaches Considered](#52-alternative-approaches-considered)
+    - [5.3 Comparison with Existing PropAMM Deployments](#53-comparison-with-existing-propamm-deployments)
+  - [6. Backwards Compatibility](#6-backwards-compatibility)
+  - [7. Test Cases](#7-test-cases)
+  - [8. Security Considerations](#8-security-considerations)
+  - [9. License](#9-license)
+
+## 1. Summary
+
+This BNB Chain Application Proposal (BAP) introduces the **Unified PropAMM Relayer**, a builder-neutral professional liquidity execution layer for BNB Chain, together with the unified on-chain **PropAMM Standard** it executes against. Makers update quotes at high frequency through a single Relayer API. Wallets, aggregators and solvers keep using standard EVM JSON-RPC (`eth_call`, `debug_traceCall`, `eth_sendRawTransaction`). The Relayer simulates, matches, bundles and broadcasts atomic bundles to all connected builders.
+
+> **Maker sends quotes. Taker sends transactions. Relayer builds bundles.**
+
+## 2. Abstract
+
+A proprietary AMM (PropAMM) is a market-maker operated pool whose pricing is driven by off-chain quotes rather than by an on-chain curve alone. PropAMMs deliver tighter spreads and better fills than passive AMMs, but every existing deployment pushes the execution complexity onto one participant: makers on Solana run chain-specific landing infrastructure, makers on Base pay to publish every quote update on-chain, and takers on Titan consume a builder-specific state override stream and construct bundles themselves.
+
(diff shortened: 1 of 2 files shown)


nash-takapu:
Thank you @asiawildboar for proposing BAP-710 and for putting together the Unified PropAMM Relayer design. The proposal provides a systematic approach to PropAMM quote updates, transaction simulation, and Builder execution, and attempts to improve quote inclusion efficiency.

After carefully reviewing the proposal, we have the following concerns.

### Taker

- **Reduced transaction composability:** Transactions involving PropAMM must be sent through the designated Relayer for Bundle construction. They can no longer be freely combined with other transactions, external Bundles, or other Builders, which makes the design particularly difficult to use for searchers performing copy trading, arbitrage, and other strategies that require custom Bundle construction and submission.

### Wallet

- **The integration cost is underestimated:** Although the transaction format and signing flow remain unchanged, Wallets still need custom development to distinguish PropAMM transactions from ordinary transactions and to handle different simulation, broadcasting, and failure paths.

- **Existing RPC business relationships are difficult to replace:** Wallets choose RPC providers based not only on APIs, but also on rebates, performance, reliability, multi-RPC availability, and existing commercial partnerships. Connecting to a single Relayer would affect existing RPC traffic and commercial arrangements, creating additional adoption barriers.

- **Transaction stability may suffer during network congestion:** PropAMM transactions sent directly to a Builder may fail because the corresponding Quote Update is missing. At the same time, larger Bundles assembled by the Relayer may be harder to order and include than existing single transactions during periods of congestion. This could affect Wallet transaction stability and create significant concerns about integration.

### Aggregators / Solvers

- **The main integration challenge is maintaining a low-latency in-memory pool state:** Efficient routing and liquidity splitting require the latest quote state to be available directly in memory. Querying an RPC node for every quote is not viable.

- **Both approaches require state adaptation, whether subscribing to State Override updates or running a Local Relayer:** With State Override subscriptions, the Solver must receive updates in real time and convert and apply them to its own in-memory pool state. Running a Local Relayer only moves state distribution and aggregation locally; the Solver still needs to convert the effective state maintained by the Relayer into the representation used by its own in-memory pool. Therefore, neither approach eliminates the need for state-model adaptation; they only determine where that adaptation takes place.

### Market Makers

- **The incremental benefit may not justify the migration cost:** There are already multiple PropAMMs on BSC, and major Makers and aggregators already support multi-route, many-to-many integrations. The existing ecosystem is broadly functional. The main incremental benefit of this proposal may be faster quote inclusion, but the expected improvement appears limited, while Makers would need to migrate to a unified contract and storage design that may restrict differentiation in contract and pricing logic.

- **Quote execution becomes overly dependent on the Relayer:** Makers would need to trust the Relayer to submit their quotes faithfully. The Relayer could incorrectly or maliciously submit already-signed quote updates, and software bugs could also result in wasted Gas.

- **There may be insufficient migration incentives:** It remains unclear whether existing PropAMMs would be willing to adapt and migrate. Without sufficient Maker liquidity, Wallets and Solvers would also have little incentive to integrate the system.

### Builders

- **Builders have no clear incentive to prioritize Relayer Bundles:** Builders may not need to implement PropAMM-specific logic, but the proposal does not provide a clear additional incentive for them to prioritize Relayer Bundles. Compared with existing transactions, these Bundles may be more complex and larger, and may be harder to order and include during periods of congestion.

- **The Relayer may affect fair competition among Builders:** By controlling a significant amount of transaction flow and Bundle distribution, the Relayer may influence competition and traffic allocation among Builders.

### Relayer

- **Costs and responsibilities are highly concentrated:** The Relayer must handle quote processing, state maintenance, transaction matching, Bundle construction, multi-Builder distribution, as well as the development, auditing, and long-term operation of highly available infrastructure.

- **The revenue model is unclear:** The proposal does not clearly explain how the Relayer can generate sustainable revenue sufficient to cover its infrastructure, development, and maintenance costs.

- **The single point of failure is too significant:** The Relayer occupies a highly centralized position. If it experiences an outage or serious degradation, all PropAMM transactions depending on it may be affected, potentially creating significant reputational and trust-related consequences.

### Conclusion

We believe the current BSC PropAMM ecosystem is already broadly usable. The main need at this stage appears to be improving quote-update and transaction-inclusion efficiency, rather than introducing a highly centralized Unified Relayer.

This proposal requires Wallets, Solvers, Market Makers, Builders, and the Relayer to bear significant technical, commercial, and operational costs, while the expected benefits appear limited. As a result, adoption may be challenging.

There are already protocols in the ecosystem exploring innovations in contract design, quote mechanisms, and execution. We believe it would be more valuable to encourage open competition and diverse infrastructure, allowing different participants to continue innovating, rather than concentrating PropAMM transaction flow in a single operator.

blockrazor-admin:
Thank you for proposing BAP-710. At BlockRazor, we recognize the value of this approach in protecting makers and would like to share feedback based on our practical experience developing and rolling out a similar system.

We completed internal testing and launched a similar system for matching maker quote updates with taker transactions in May of this year. We have also engaged with major wallets, swap routers, and other participants on BNB Chain, and some wallets and routers have integrated with our test system. During this process, we encountered the following challenges:

1. **Incentives for routers to integrate.** When we spoke with routers in June, most had not yet completed their PropAMM integrations. From a technical perspective, integrating with this system is relatively straightforward. However, now that most routers have integrated directly with PropAMMs through on-chain interactions, the practical question is whether the incremental benefits justify supporting an additional quoting and execution path. A low integration cost does not necessarily provide a compelling business reason to adopt the system.

2. **Wallet dependency on a single service and the impact on user experience.** This architecture ties pre-submission simulation and the submission of quote-dependent transactions to the Relayer, making it a critical dependency for execution. Without an alternative execution path, a service outage would prevent dependent PropAMM swaps from completing. If this model were widely adopted across the chain, the impact of an outage would grow accordingly, significantly affecting wallet users and the availability of on-chain trading.

   In addition, if a wallet requires that user swaps must not be included on-chain if they would revert, the Relayer must coordinate bundle simulation, replacement, and invalidation while quotes are continuously updated, on-chain state changes, and multiple builders process submissions in parallel. This substantially increases the complexity of bundle construction and execution management.

   This matching model also relies heavily on bundles: a user transaction must be included together with its corresponding quote updates. Compared with submitting a private transaction or broadcasting directly to the mempool, these additional quote dependencies and atomic inclusion constraints could significantly increase inclusion latency during network congestion, further affecting the user experience. We therefore believe that failover, mechanisms to prevent reverting transactions from being included, and swap success rates and inclusion latency under congestion all need to be clearly designed and validated before broader adoption.

3. **The complexity of supporting multiple quoting paths for makers.** Makers may need to retain their existing approach of submitting quote updates directly to builders or the mempool while also submitting quotes through the Relayer. Running both paths in parallel introduces differences in quote visibility, update timing, and execution conditions, with substantial implications for quoting strategies and risk management. These adaptation costs should also be considered when evaluating the proposal.

These are practical challenges and feedback we encountered during our rollout. We believe this model can offer stronger maker protection and, in theory, enable more competitive quotes. However, it remains difficult to quantify how much pricing improvement these protections would actually deliver. Without verifiable evidence of the benefits, it is difficult to persuade wallets, routers, and makers to bear the costs of changing their existing workflows. We recommend comparative testing that measures spreads, execution prices, success rates, and transaction latency to give participants a clearer basis for evaluating adoption.

Additionally, the Relayer handles sensitive maker quotes and transaction matching, giving its operator considerable influence over quote usage and transaction processing. We recommend making **a Trusted Execution Environment (TEE) a requirement for the Relayer’s quote processing and matching logic**, combined with remote attestation and auditable execution rules. This would allow participants to verify the logic actually being executed and reduce the risk of operators or other third parties manipulating matching for commercial or other reasons at users’ expense. A TEE alone cannot address service availability or all forms of censorship, so it should be complemented by explicit matching rules, failover mechanisms, and support for multiple relayers.

We look forward to seeing the proposal further address integration incentives, execution availability, and trust in the matching process alongside maker protection, helping establish the conditions for adoption at scale across the BNB Chain ecosystem.

PR #716, opened 2026-09-17
BEP-702: creation log order, bootstrap caller, sentinel planting, event attribution

zlacfzy:
BEP-702 follow-ups from the fix review, matched by the bsc implementation PR (linked in comments).

* **§3.4 `createCAS20` order**: `CAS20Created` is emitted as soon as the token's identity is written, before the initial admin's `RoleGranted` and any `initCalls`.
* **§3.4 bootstrap window**: inside `initCalls` `msg.sender` is the factory, as for any contract the factory calls; methods that act on `msg.sender` have no place in the window.
* **§3.15 / §3.16 sentinels (FR-01)**: the registries' sentinels are planted only on the block that crosses into Jenner. Every BSC network schedules `jennerTime` after its genesis and a new network must do the same; the genesis-allocation requirement, which no implementation performs, is withdrawn.
* **§3.4 `CAS20Created` attribution (FR-03)**: the event carries neither creator nor salt; the creator is the `createCAS20` caller and `getCAS20Address` verifies a (creator, salt) claim against the token address.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


zlacfzy:
Implementation: bnb-chain/bsc#3826

will-2012 (approved):

